PR #107, opened 2014-10-24
"NGHz  " displayed as "N,00GHz"

ukasz:
This way 1GHz will be showed up as a 1,00Ghz, so if you have cpu with load based frequency scalling displayed value will be displayed with close to constant length. Before that we had "1  ", "800","1,70" with this change it will be "1,00","800","1,70" - no trailing spaces after frequency, when frequency is divisble by 1GHz.


lukefromdc:
Any updates on this or can we close it? Last activity was over 6 years ago, when we were on GTK 2

PR #146, opened 2015-07-01
charpick, geyes: restore icons on buttons in prefs dialogs

monsta:
Removing something that users see and interact with only to fix some "deprecations" wasn't right. Let's get the icons back.


NiceandGently:
nice, this brings the missing icons in prefs dialogs back

infirit:
I don't remember those commits exactly but loosing the icons was not my intention.

monsta:
No problem, it's solved now :smile:

PR #161, opened 2015-10-15
Bugfixes and theme support

lukefromdc:
Drivemount: support custom gtk3 panel themes, robust support for transparent themes
Charpick: support custom gtk3 panel themes, robust support for transparent themes
Cpufreq: support custom gtk3 panel themes, robust support for transparent themes
Cpufreq: fix gtk3 rendering issues with 0px wide labels and bottom panel rendering. Someone should check to ensure label width fix does not cause gtk2 issues but the direct numerical width specification should work there too. This one committed last in case issues arise in gtk2 so it can be split out

### DIFF
--- a/charpick/charpick.c
+++ b/charpick/charpick.c
@@ -319,6 +319,20 @@ populate_menu (charpick_data *curr_data)
 		list = g_list_next (list);
 	}
 	build_table(curr_data);
+	
+	/*Set up custom theme and transparency support*/
+#if GTK_CHECK_VERSION (3, 0, 0) 
+	GtkWidget *toplevel = gtk_widget_get_toplevel (menu);
+	/* Fix any failures of compiz/other wm's to communicate with gtk for transparency */
+	GdkScreen *screen = gtk_widget_get_screen(GTK_WIDGET(toplevel));
+	GdkVisual *visual = gdk_screen_get_rgba_visual(screen);
+	gtk_widget_set_visual(GTK_WIDGET(toplevel), visual); 
+	/* Set menu and it's toplevel window to follow panel theme */
+	GtkStyleContext *context;
+	context = gtk_widget_get_style_context (GTK_WIDGET(toplevel));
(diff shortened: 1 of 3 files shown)


lukefromdc:
I was able to merge Ravit65's commit into a branch of the branch this PR was taken from, yielding this commit that would update the PR to what is otherwise current. This was done on a branch so as not to alter the PR itself

https://github.com/lukefromdc/mate-applets/commit/8cf550a774a53d1f8aa654db53ea855652daa801

raveit65:
np :)

raveit65:
merged
https://github.com/mate-desktop/mate-applets/commit/ec37a4ececaecd61794913469b82a301919f8a08
https://github.com/mate-desktop/mate-applets/commit/167c7c5652cc62e3346f5238e15acb3b4b0bcf9b
https://github.com/mate-desktop/mate-applets/commit/cd971d39fcafc1aa71df342060b039012f1e05af
https://github.com/mate-desktop/mate-applets/commit/7b3bf0dc42c752c4403e1fec4778e42e58914388
Thank you

raveit65:
I added gtk3 condition for https://github.com/mate-desktop/mate-applets/commit/ec37a4ececaecd61794913469b82a301919f8a08
It seems to be working with gtk2 too, but we don't want to intoduce any possible regression for gtk2 during feature freeze for 1.12.

raveit65:
btw. Do you have any idea why orientation on vertical panel does not work with cpufreq applet?
Here the label and label_unit are always left from icon, with gtk2 this is under the icon.
As you know i could fix that for horizontal panel.

lukefromdc:
I do not know why it does this, other than it seems that in gtk3 whatever sets it to
go underneath on vertical panels  in gtk2 does not work. What I had to do for label width
 may offer a clue: the code in gtk2 that computes the label width also  does not work in gtk3, 
thus I had to by experiment find values that would work for hardcoding the width of the labels
while using as little space as possible.

PR #166, opened 2015-11-18
drivemount: Don't remove background

City-busz:
This fixes the invisible text for GTK+ 3 themes which don't support mate-panel explicitly. Themes could still overwrite the background if they want in the usual way.


lukefromdc:
I just tested this with my theme, since it sets the background explicitly this does not require any changes on my theme or presumably any other that sets a custom background. This is good to go as far as I am concerned, and hopefully will work wherever else this problem turns up.

PR #172, opened 2015-12-26
gtk3: stop cpufreq-applet turning transparent 

lukefromdc:
gtk3: stop cpufreq-applet turning transparent -prevent the cpufreq applet and panel area beneath from becoming transparent to the desktop when a menu item in the popup menu is clicked. Verified to work on gtk3.19.5. Does anyone know a better way to do this?


raveit65:
confirmed with 3.16

raveit65:
Thank you,
merged https://github.com/mate-desktop/mate-applets/commit/4dbb7347a4b344658d28cfd49f6a925f000422ee

monsta:
Interesting... looks like the ```parent``` argument isn't even used in these two ```cpufreq_selector_set_xxx_async``` functions. It gets assigned to a field of struct, but no one reads that field afterwards. Maybe just remove it? :smile:

lukefromdc:
new PR to remove "parent":
https://github.com/mate-desktop/mate-applets/pull/177
works fine in gtk3.19.5, needs test in GTK2

PR #182, opened 2016-03-07
cpufreq  GTK3: hardcode fewer labels

lukefromdc:
In gtk3,  replacing "label" with "applet->label" makes the gtk_widget_get_preferred_width functions work for the frequency and unit labels without issue. Trim them a bit in gtk3.19 builds as they render 2px wider there for some reason. Two out of three options for text now work as they did in GTK2. 

The exception is the percentage label. That one apparently starts up with "0%" entered and does not seem to update its width when larger character strings come up.  This even though the label itself can be shown to be changing it's size with the length of the string by invoking the GTK 3.19 applet transparency issues with the label themed black. Thus, a fixed figure of 20px still has to be added to that one. On the other hand, that will ensure the applet never jumps from percentage changes. Height of the whole applet is still limited somewhere to 24 px, still looking for a fix for that too.

### DIFF
--- a/cpufreq/src/cpufreq-applet.c
+++ b/cpufreq/src/cpufreq-applet.c
@@ -16,7 +16,7 @@
  *  License along with this library; if not, write to the Free
  *  Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
  *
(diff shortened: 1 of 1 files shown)


raveit65:
Nice :)
as a side effect you fixed wrong orientation of labels/units on vertical panels.

PR #186, opened 2016-03-30
GTK 3.20 cpufreq: fix themes clipping units label

lukefromdc:
Stop BlackMATE, possibly other themes from cutting off the end of the units label and get rid of an #if block for GTK 3.19 at the same time.  GTK 3.19 seemed to pad the end of this label a bit, not so with released GTK 3.20.

### DIFF
--- a/cpufreq/src/cpufreq-applet.c
+++ b/cpufreq/src/cpufreq-applet.c
@@ -334,12 +334,7 @@ cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 		label = gtk_label_new (freq_text);
 #if GTK_CHECK_VERSION (3, 0, 0)
 		gtk_widget_get_preferred_width (applet->label, &label_width, NULL);
-
-#if GTK_CHECK_VERSION (3, 19, 0)
-		width = MAX (width, label_width)-2;
-#else
-		width = MAX (width, label_width);
-#endif
+		width = MAX (width, label_width); 
 #endif
 #if !GTK_CHECK_VERSION (3, 0, 0)
 		gtk_widget_size_request (label, &req);


raveit65:
Hmm, with the PR the label is cut off for all themes on horizontal panel, include BlackMate from latest git version.
cpu-freq
And without this PR but with latest BlackMate changes.
cpu-freq-2

All with gtk3-3.20.1-1.fc24.x86_64.

lukefromdc:
Which GTK version, which label, whch theme? . I can easily pad them a bit. 
I had trouble in BlackMATE with non-zero padding in the theme causing the
labels to not appear at all on restarting the panel. I only tested on a horizontal
panel, but the old hardcoded labels caused issues with vertical panels.

There should only be a change on GTK 3.20, in which code to narrow the label by
2px was removed. GTK 3.18 and earlier should be exactly the same built code.

If the old code with the 2px subtracted prevented some bug with the panel cutting
the label, we can close this PR and try something else for BlackMATE. No other
theme I have cut the last 2PX off the label with the old code, BlackMATE was 
showing (both versions) that it was possible to do so. If this is still happening I
need to reverse the former subtraction and add a couple px to that label but not
enough to mess up vertical panel labels.

raveit65:
1. I use gtk3-3.20.1-1.fc24.x86_64.
2. I use latest mate-themes from gtk-3.20 branch with your commit for BlackMate.
And in my test i see the cut off the labels only if i build latest mate-applets from git + this PR,
with every theme from mate-themes.
If i use a tarball from latest git without your PR everything is fine here.
I will re-check everything again , maybe i did a mistake with building or what ever.
And the issue with freq applet on vertical panel was gone after your previous PR for the applet.

This is the tarball which include your PR and breaks the labels for me, please test.
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/SOURCE/mate-applets-1.13.1.tar.xz

raveit65:
> If the old code with the 2px subtracted prevented some bug with the panel cutting the label, we can close this PR and try something else for BlackMATE.

I don't think that is needed, here latest BlackMate works fine.

lukefromdc:
OK, I just tested your tarball with GTk 3.20.1 from the GNOME tarball as released.
with BlackMATE, with Green-Submarine, and with my own theme. The only thing 
that clipped the units label was changing to a theme that hardcodes a larger text
size (my own theme), because the size had already been allocated. Doing so would
cut off all of the last character. Restarting the panel fixed that every time. I could not
by any other means get the label to cut off.

Actually, the current themes entirely removed all the trouble for me with clipped labels
except for the changing text size issue, in which case it appears the allocation is not
being recomputed until the panel is restarted. Possibly that was also why having run the
panel with another theme kept the label alive in BlackMATE with the previous buggy
code.

Prior to yesterday's change in BlackMATE, starting the panel with BlackMATE as the
therme would prevent the entire units label from appearing at all. It would appear-but
clipped 2px short, appearing only if the panel was started from another theme and 
changed to BlackMATE. Removing the code to clip 2px from the label in GTK 3.19 or
later builds only did not fix the whole label missing problem but did make it render all
the way to the end of the label on a switch into BlackMATE. Thus the theme itself
also needed work, the 1px default padding turned out to be the cause of the whole
units label not coming up.

We can keep this or close it. If the old code is preventing ANYONE from getting a bug,
I vote to keep it until the root cause of that bug is found. If the bug came from a build
mistake, there is the option to remove an #1f-#endif block and use slightly simper code.

My build options were (schemas do not change with this PR):

./autogen.sh --with-gtk=3.0 --prefix=/usr  --sysconfdir=/etc --enable-shared=yes --disable-maintainer-mode  --disable-schemas-compile

raveit65:
Ok, it works now after creating a new tarball and rebuilding the rpm.
Tested with 3.20.1 and 3.16.x, i don't have a gtk+-3.18.x installation on bare metal to test it there.
And yes. i noticed a cut of with themes which using bold fonts, (contrastHighInverse gtk+-3.16)
I will merge it.
Thank you for the patient with me :)

lukefromdc:
More testing is always good, and it pays to fix things right the first time when possible,
less effort than a commit followed by another to fix bugs from the first one. To me, this
is about trying to squash as many as possible of the bugs that make a DE look
"unprofessional," the papercuts that a random user would consider cumulative demerits.

I don't know if we have a big enough team to really catch up as fast as GTK changes on
us, but it would be really nice to have MATE with GTK 3 working at least as well as 
GNOME 2 did and as widely distributed. Many are saying MATE with GTK 3 has more 
potential than Cinnamon does, being lighter, faster code.

raveit65:
> I don't know if we have a big enough team to really catch up as fast as GTK changes on us, but it would be really nice to have MATE with GTK 3 working at least as well as GNOME 2 did and as widely distributed. Many are saying MATE with GTK 3 has more potential than Cinnamon does, being lighter, faster code.

Yes, our small team is a bit the problem, it would be nice if 2 or 3 people more get involved.
We will see.
Currently the biggest blocker for me is not in our hand, hope that cairo guys will react on the report. https://bugs.freedesktop.org/show_bug.cgi?id=91967
Beside from this we should think about if we really need broken background image support for caja, or color selecting in mate-appearance-properties.
All this is droped in gnome, so no chance to back port something. And i'm really in doubt that we can write the code for ourselves.
But beside from those issues we can do a switch to gtk3.
Well, there are some more minor issues, but those should not stop the transition to gtk3.
And gnome was also far a way from perfect with 3.0 release.
I forgot, pluma and eom python plugins is pretty broken.
Sadly, i'm not really a coder, i'm good in themes and back porting stuff, but really writing code is to much for me.

lukefromdc:
GNOME 2 could be quite buggy at times, my sister had a laptop with Ubuntu Lucid
on it. Almost every startup included one or more crashing panel applets, plus she
was often stumped by launchers and even applets appearing and disappearing.

lukefromdc:
That Cairo bug is years old, a fix offered, and no action taken. Any way to work 
around it, say by handling Cairo's error in such a way as not to take out the
whole process? Some way to "sandbox" this?

lukefromdc:
I just found a bug that can cause labels to be clipped on certain machines only. If the panel is started while the CPU governor is in its lowest state, it's possible for the frequency to momentarily be reported as zero. This causes the space for the frequency label to be allocated for one character only, when the labels are rendered the missing space is clipped from the last label shown, the units label.  If percentage is set instead of frequency I've occasionally seen zero percent shown at applet startup, verifying this.

My guess is a firmware bug causes the momentary reporting of zero as I cannot duplicate this on a machine I have locally with the same board but  a different BIOS version than the affected machine. 
I suspect this may always be the case but the speed at which it catches up varies with the board and BIOS, as I've see that zero percent shown for a moment on machines that never clip labels when frequency and unit are selected for display.

On boot, the CPU governor is always initially at max clocks so this only shows up on a new session or a restarted panel. Setting the governor back to a setting other than the lowest, then restarting the panel will show the whole label.

Fix for this (assuming no backend fix) would be to get the applet space to recompute allocation each time the frequency changes, and it should also be recomputed when the text size changes to prevent larger text from clipping labels. Not sure how to rerun the functions in question for this.

PR #188, opened 2016-04-25
Update backend.c

alejandro-perez:
Actually, a default route is defined by having the destination subnet to 0.0.0.0/0, not having the GW to 0 (which in any case, should be gw != 0).
I just discovered this bug when starting docker and having many interfaces without GWs.
With the proposed change it works fine with me.

### DIFF
--- a/netspeed/src/backend.c
+++ b/netspeed/src/backend.c
@@ -109,7 +109,7 @@ get_default_route(void)
 
 		if (retval != 11) continue;
 
-		if (gw == 0 && !is_dummy_device(device)) {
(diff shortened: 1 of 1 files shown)


raveit65:
Thank you for picking up an old missing fix from gnome-applet-network
merged
https://github.com/mate-desktop/mate-applets/commit/4467ec4c53c8aa09ba744bb7f4f1be76773ed892

alejandro-perez:
You're welcome :) Great job you are doing with MATE. I'll try to contribute as much as I can.

Issue #193, opened 2016-06-08
cpufreq build failed with kernel-4.7


raveit65:
```
cpufreq-monitor-libcpufreq.c: In function 'cpufreq_monitor_libcpufreq_run':
cpufreq-monitor-libcpufreq.c:114:8: warning: implicit declaration of function 'cpufreq_cpu_exists' [-Wimplicit-function-declaration]
   if (!cpufreq_cpu_exists (cpu)) {
        ^~~~~~~~~~~~~~~~~~
/bin/sh ../../libtool  --tag=CC   --mode=link gcc  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic  -Wl,-z,relro -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -o mate-cpufreq-applet cpufreq-applet.o cpufreq-utils.o cpufreq-prefs.o cpufreq-selector.o cpufreq-popup.o cpufreq-monitor.o cpufreq-monitor-factory.o cpufreq-monitor-procfs.o cpufreq-monitor-sysfs.o cpufreq-monitor-libcpufreq.o cpufreq-monitor-cpuinfo.o -lmate-panel-applet-4 -lgtk-3 -lgdk-3 -lpangocairo-1.0 -lpango-1.0 -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lmate-desktop-2 -lgtk-3 -lgdk-3 -lpangocairo-1.0 -lpango-1.0 -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lstartup-notification-1 -lcpupower -ldbus-glib-1 -ldbus-1 -lgobject-2.0 -lglib-2.0 
[...]

PR #194, opened 2016-06-09
fix build with kernel-4.7

leigh123linux:
Should fix https://github.com/mate-desktop/mate-applets/issues/193

### DIFF
--- a/cpufreq/src/cpufreq-monitor-libcpufreq.c
+++ b/cpufreq/src/cpufreq-monitor-libcpufreq.c
@@ -23,8 +23,11 @@
 #include <glib/gi18n.h>
 
 #include <stdlib.h>
+#include <linux/version.h>
 #include <cpufreq.h>
-
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 7, 0)
+#include <cpuidle.h>
+#endif
 #include "cpufreq-monitor-libcpufreq.h"
 #include "cpufreq-utils.h"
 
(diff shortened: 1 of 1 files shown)


raveit65:
PR works fine with building against kernel 4.6 and latest 4.7.0.rc2
@monsta @flexiondotorg 
Do we need a require for kernel-headers in configure?
For fedora build system this isn't needed, it is always there.
But i don't know the situation in debian & co and others.

leigh123linux:
looking at the existing headers it already requires it to build

```
$ rpm -qf /usr/include/cpufreq.h
kernel-headers-4.6.2-1.fc24.x86_64
kernel-tools-libs-devel-4.5.5-300.fc24.x86_64
```

monsta:
Did you guys test this with ```--with-cpufreq-lib=cpufreq``` option as well?

leigh123linux:
@monsta I'm not able to test that as /usr/lib/libcpufreq.so and /usr/lib64/libcpufreq.so don't exist on fedora.

```
libtool: link: gcc -g -O2 -Wl,-z -Wl,now -o mate-cpufreq-selector cpufreq-selector.o cpufreq-selector-sysfs.o cpufreq-selector-procfs.o cpufreq-selector-libcpufreq.o cpufreq-selector-service.o cpufreq-selector-factory.o main.o  -lcpufreq -lpolkit-gobject-1 -lgio-2.0 -ldbus-glib-1 -ldbus-1 -lgobject-2.0 -lglib-2.0
/usr/bin/ld: cannot find -lcpufreq
```

I don't expect any issues if the distro has libcpufreq.so installed

http://packages.ubuntu.com/precise/amd64/libcpufreq-dev/filelist

monsta:
@flexiondotorg: I think we need to keep an eye on this for future Debian/Ubuntu releases...

raveit65:
@monsta same for me, i can't test this in fedora
For the record, this select only the current kernel during build time.
This means if your distro switch to kernel-4.7 you need to rebuild the package.
And it does not change the function in runtime, ie. if a user or admin use a lower kernel version.

Issue #196, opened 2016-07-01
eyes applet does not respect panel orientation


lyssdod:
Hello,

I'm using vertical (left) panel orientation, and "bizarre" eyes theme. With it eyes look like this:
![left](https://cloud.githubusercontent.com/assets/2615950/16536688/96b534d2-3ffd-11e6-8d6e-a149eab4168d.png)

PR #198, opened 2016-07-02
Make eyes respect panel orientation

lyssdod:
Small and obvious fix for #196 

Before           |  After
:-------------------------:|:-------------------------:
  |

### DIFF
--- a/geyes/geyes.c
+++ b/geyes/geyes.c
@@ -225,8 +225,8 @@ setup_eyes (EyesApplet *eyes_applet)
 {
 	int i;
 
-        eyes_applet->hbox = gtk_hbox_new (FALSE, 0);
-        gtk_box_pack_start (GTK_BOX (eyes_applet->vbox), eyes_applet->hbox, TRUE, TRUE, 0);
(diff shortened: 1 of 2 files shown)


monsta:
I've tested it with the default eyes theme, and it looks weird...

panel is 24 px wide:
eyes-24

panel is 48 px wide:
eyes-48

panel is 72 px wide:
eyes-72

So, they're placed diagonally, not vertically, and seem to be aligned to panel edges each other...

lyssdod:
Yes, interesting...How should they be played then? Centered? Looks like some default alignment for gtk_vbox

monsta:
Well, I don't know, but diagonal placing and this weird alignment don't look right for sure.

monsta:
I'll close this for now, since it apparently needs more work before it can be merged.

lyssdod:
I would like to get it merged finally (after fixing weird placing), thus I would like to discuss possible workarounds of the issue:
1. Stick eyes to center line?
2. Try to scale them to fill all the space?
3. Any other option?..

PR #201, opened 2016-07-31
drop dependency to libmate-deskop

raveit65:
Using GtkColorButton or GtkColorChooser makes it possible to get rid of deps to libmate-desktop.
MateColorButton was an implementation from Stefano to make my life easier for themes with using a own mate-color-chooser.
But using this lib for an app which set only the line color for a graph does not make any sense.
Here using gtk stuff is sufficient.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -15,7 +15,6 @@ GLIB_REQUIRED=2.36.0
 GIO_REQUIRED=2.15.3
 LIBPANEL4_REQUIRED=1.7.0
 LIBGTOP_REQUIRED=2.11.92
-LIBMATE_DESKTOP_REQUIRED=1.9.0
 LIBNOTIFY_REQUIRED=0.7.0
 UPOWER_REQUIRED=0.9.4
 DBUS_REQUIRED=1.1.2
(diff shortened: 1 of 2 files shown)


monsta:
Yeah, I was thinking about this too... Looked at all our packages, found out that less than half of them use MateColorButton. So users should already know how the color select dialog looks in modern GTK+3.

monsta:
All good in GTK+2 build.

monsta:
New color button & dialog work fine in GTK+3 build.

monsta:
However, I wonder how we'll handle possible future upgrade from GTK+2 to GTK+3 build... the format in which colors are stored in GSettings is completely different.

raveit65:
yes, gtk3 use rgb value or rgba id alpha color is enabled, and gtk2 use hex values.
Last commit changed that because using hex here did not work.
```
Previous commit writes garbage to gsettings, in result no graphs are displayed
with a new start of the applet.
```
So it's clear that a old gsetting value won't read, i hope that in this case the default values in code will use.
Otherwise we have to add something like this.
If 'gesetting value is gabage' than  'use defaut value'
Or some code which converts old hex values.

Thanks for testing, i will merge it.

monsta:
Looks like mate-terminal still uses the old format in GSettings...

PR #202, opened 2016-08-04
deprecations

raveit65:
Please review.
Note, i noticed an issue with sticky-notes, but it is unrelated to the commit here, happens with master too.
Changing the font color will only apply after a restart of s-n (gtk3).

### DIFF
--- a/battstat/battstat_applet.c
+++ b/battstat/battstat_applet.c
@@ -1317,6 +1317,49 @@ load_preferences(ProgressData *battstat)
   battstat->showtext = g_settings_get_int (settings, "show-text");
 }
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+/* Convenience function to attach a child widget to a GtkGrid in the
+   position indicated by 'loc'.  This is very special-purpose for 3x3
+   gridss and only supports positions that are used in this applet.
+ */
+static void
+grid_layout_attach (GtkGrid *grid, LayoutLocation loc, GtkWidget *child)
+{
+  switch( loc )
+  {
+    case LAYOUT_LONG:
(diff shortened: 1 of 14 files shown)


lukefromdc:
Tested with GTK 3.21 and seems to work fine here. Panel has menubutton, desktop, launchers, window-list, inhibit, force-quit, window-switcher, hardware sensors, CpuFreq,tray, clock, trash, quit in that order. Tested geyes, fish(not from applets package), weather (not configured), search, invest in another panel, no obvious issues.

battstat/battstat_applet.c
```diff
@@ -1535,13 +1604,24 @@ create_layout(ProgressData *battstat)
...
+#else
   /* Let reconfigure_layout know that the table is currently empty. */
+#endif
   battstat->layout.status = LAYOUT_NONE;
   battstat->layout.text = LAYOUT_NONE;
   battstat->layout.battery = LAYOUT_NONE;
 
+#if GTK_CHECK_VERSION (3, 0, 0)
+  /* Put the grid directly inside the applet and show everything. */
+  gtk_widget_set_halign (battstat->grid, GTK_ALIGN_CENTER);
```
monsta: Hmm is that needed? I don't see the corresponding code for GTK+2...

monsta:
GTK+2 build seems ok.

monsta:
> Changing the font color will only apply after a restart of s-n (gtk3).

Just experimented a bit with both global notes options and properties of the existing note... couldn't reproduce that. Font color and size are changing as they should.

monsta:
Looks ok in GTK+3 as well.

raveit65:
all merged!

PR #217, opened 2016-11-19
build with kernel 4.8

dnk:


raveit65:
This breaks every build with kernel-4.8 for me.

monsta:
Well, this builds in Ubuntu 16.10 with kernel 4.8 and ```--with-cpufreq-lib=cpufreq``` build option. I guess it would break when libcpupower is chosen instead of libcpufreq for build... But I can't test it, Ubuntu still doesn't have a proper dev package for libcpupower (https://bugs.launchpad.net/bugs/1215411).

This bug is really annoying, we need a good solution for both old and new kernels and for both libcpufreq and libcpupower...

raveit65:
> Well, this builds in Ubuntu 16.10 with kernel 4.8 and --with-cpufreq-lib=cpufreq build option. I guess it would break when libcpupower is chosen instead of libcpufreq for build

But is this really needed?
Btw. is cpu-freq applet build broken in ubuntu without this mystical PR? 
I can't remember that @flexiondotorg ever mentioned a problem with building the applet.
And as always @dnk never added a valid description to his PR's, ie. in which distro this is a problem ;-)

flexiondotorg:
I've added a patch to the Debian and Ubuntu packages, like the one @monsta included, so that mate-applets builds.

monsta:
There's [this patch](https://sources.debian.net/src/mate-applets/1.16.0-1/debian/patches/1000-fix-build-on-linux-4.7-or-newer.patch/) by @flexiondotorg which is used in Debian and Ubuntu. Obviously we can't use it upstream, it breaks libcpupower support.

monsta:
There was also a discussion at https://github.com/mate-desktop/mate-applets/commit/a708224decae0f7417aa38b926318fa32747b161.

monsta:
As mentioned in https://github.com/mate-desktop/mate-applets/issues/238, `HAVE_LIBCPUFREQ` is defined simply when `cpufreq.h` is found, so it doesn't really help.

PR #227, opened 2017-02-07
drivemount: fix runtime warnings with GTK+ 3.20/22

monsta:
Now I don't know if I do it right, because this time these CSS settings are in class init function, and in fact DriveButton is inherited from GtkButton. So maybe the right thing is to set a class name, not a widget name. But it seems to work anyway, and ```gtk_widget_class_set_css_name``` is only available for GTK+ >= 3.20, while I need to make it work for older versions too. Feel free to make any corrections. :slightly_smiling_face:

### DIFF
--- a/drivemount/drive-button.c
+++ b/drivemount/drive-button.c
@@ -71,13 +71,25 @@ drive_button_class_init (DriveButtonClass *class)
     GtkCssProvider *provider;
 
     provider = gtk_css_provider_new ();
+
+#if GTK_CHECK_VERSION (3, 20, 0)
     gtk_css_provider_load_from_data (provider,
-                                     "DriveButton {\n"
+                                     "#drive-button {\n"
+                                     " border-width: 0px;\n"
+                                     " padding: 0px;\n"
+                                     " margin: 0px;\n"
+                                     "}",
+                                     -1, NULL);
+#else
+    gtk_css_provider_load_from_data (provider,
+                                     "#drive-button {\n"
                                      " border-width: 0px;\n"
                                      " -GtkWidget-focus-line-width: 0px;\n"
                                      " -GtkWidget-focus-padding: 0px;\n"
                                      "}",
                                      -1, NULL);
+#endif
+
     gtk_style_context_add_provider_for_screen (gdk_screen_get_default(),
                                     GTK_STYLE_PROVIDER (provider),
                                     GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
@@ -99,6 +111,8 @@ drive_button_init (DriveButton *self)
     self->update_tag = 0;
 
     self->popup_menu = NULL;
+
+    gtk_widget_set_name (GTK_WIDGET (self), "drive-button");
 }
 
 GtkWidget *


raveit65:
I see the same with behaviour with Menta themes too, so i think themes are OK.
I only wonder why the button don't goes smaller with unmount.

raveit65:
And i see that with adwaita theme, for some reasons the applet needs to restart if you change theme to see that effect on the icon.
So, i guess the behaviour is OK with Submarines and Menta themes.
I don't know why i don't see that with Traditional or BlackMate.
Any way, all this isn't related to PR , hopefully.

monsta:
It's actually weird that icon size changes at all. It should always be the same because the margin and padding are hardcoded to zero now. At least it should be so with GTK+ 3.20/22. With 3.18, I saw that the padding and margin are still affected by the themes that set them (Submarine themes, Menta). But I guess you tested it with 3.20 or 3.22?

raveit65:
Yes, with 3.20.
As i see this with Adwaita too i guess it is a relative new setting for images in the main css.
Only our Menta and Submarine themes use all hotest and newest stuff ;-)

monsta:
Can you check if this effect happens with earlier commits (or even in 1.16 releases)?

raveit65:
drive-mount-1 16

raveit65:
It look like that self->icon_size = 24; from drive_button_init (DriveButton *self)
https://github.com/mate-desktop/mate-applets/blob/master/drivemount/drive-button.c#L110
is only used if i change the gtk+ theme.
After a panel restart it use 16px icon from theme which is to small, imo.

lukefromdc:
I was unable to duplicate the icon size changing issue with this commit applied in any theme I tested. Icon size can vary with the icon theme and in my tests changed promptly with the icon theme change and never required a restart

raveit65:
@monsta @lukefromdc 
Is it really needed to set border-width to zero?
https://github.com/mate-desktop/mate-applets/pull/227/commits/ac0ae423c0f5d3f9e7aa4b7e473cade399d2d97d#diff-544b3212ff0efd7d311a7293925a41fdR78
Because Traditional themes use a border for applets GtkButtons. Without a border using an background for the GtkButton is pretty useless for normal button state.
And i need to remove that from theme settings for drive-mount applet.

monsta:
I mentioned that in https://github.com/mate-desktop/mate-applets/pull/226. I used it to fix the negative allocation warning, similar to earlier fixes by @lukefromdc. I don't know if there's another solution...

lukefromdc:
Oddly I got a border in my UbuntuStudio_Legacy theme even with this, and without the allocation warning. Had to explicitly remove it, it was set for buttons in general as a border image and apparently was getting rendered even with the width set to zero. It should not show up with a width of zero but did anyway, which is probably a bug in GTK3.

raveit65:
Well, personal i don't use traditional themes , but a lot of users do that.
I only want to mention that this is really a change of look and feel of the panel if a theme use borders or a border-image for normal button state, and i need to remove that.
Maybe someone with more knowledge than me can file out a report at gtk+.
Also adding a FIXME in code would be great.

raveit65:
Hopfuly we don't need to do that for tasklist buttons.

raveit65:
I found a solution :-)
```
.mate-panel-menu-bar #PanelApplet #drive-button,
.mate-panel-menu-bar #PanelApplet #drive-button:hover,
.mate-panel-menu-bar #PanelApplet #charpick-applet-button,
.mate-panel-menu-bar #PanelApplet #showdesktop-button:checked {
    border-image: url("img/border.svg") 3 3 4 3 / 3px 3px 4px 3px repeat;
    border-width: 0px;
}
```
Using a border-image and border-image-width is possible.
border-width: 0px is for making the theme compatible with < mate-1.18.
Did i miss an applet?

PR #229, opened 2017-02-23
drivemount: don't connect to change_background signal

raveit65:
taken from:
https://git.gnome.org/browse/gnome-applets/commit/?id=ea315cd

It looks like that this code isn't needed.

### DIFF
--- a/drivemount/drive-list.c
+++ b/drivemount/drive-list.c
@@ -88,7 +88,6 @@ drive_list_init (DriveList *self)
     self->orientation = GTK_ORIENTATION_HORIZONTAL;
     self->layout_tag = 0;
     self->icon_size = 24;
-    self->relief = GTK_RELIEF_NORMAL;
 
     /* listen for drive connects/disconnects, and add
      * currently connected drives. */
@@ -339,7 +338,6 @@ add_volume (DriveList *self, GVolume *volume)
 	return;
 
     button = drive_button_new (volume);
-    gtk_button_set_relief (GTK_BUTTON (button), self->relief);
     drive_button_set_size (DRIVE_BUTTON (button), self->icon_size);
     gtk_container_add (GTK_CONTAINER (self), button);
     gtk_widget_show (button);
@@ -382,7 +380,6 @@ add_mount (DriveList *self, GMount *mount)
 	return;
 
     button = drive_button_new_from_mount (mount);
-    gtk_button_set_relief (GTK_BUTTON (button), self->relief);
     drive_button_set_size (DRIVE_BUTTON (button), self->icon_size);
     gtk_container_add (GTK_CONTAINER (self), button);
     gtk_widget_show (button);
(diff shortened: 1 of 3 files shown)


monsta:
Testing with GTK+ 3.18... now the drive button doesn't follow custom panel color & transparency with TraditionalGreen theme. Might be something missing in the theme, but it worked before this PR.

lukefromdc:
Looks like this is needed after all in GTK 3.18 and earlier builds. Could exclude it from GTK 3.20 or at least 3.22 builds at the expense of more ifdefs

raveit65:
I do not have a VM with 3.18 any more as this is EOL in fedora.
But with => 3.20 traditional themes use always a background-image for
```
.mate-panel-menu-bar #PanelApplet button.flat {}
```
with or without the PR, see https://github.com/mate-desktop/mate-themes/blob/master/desktop-themes/TraditionalOk/gtk-3.0/mate-applications.css#L160   .
The background-change signal does not have any influence on GtkButton widdget , only for the applet itself.
```
.mate-panel-menu-bar #PanelApplet {}
```
For making it complicated panel settings are complete rewrote from 3.18 to 3.20 :-)
I need to back port this to 1.16 to have a chance to reproduce this in a rhel7 VM with 3.14.
Setting for the panel are here the same.
With 3.18 trad themes use always a background-image for all button states too.
```
MatePanelApplet .button {}
```
So i am still wondering that you see a transparancy GtkButton w/o this PR.

https://github.com/mate-desktop/mate-themes/blob/gtk3.18/desktop-themes/TraditionalOk/gtk-3.0/mate-applications.css#L159

raveit65:
Ok, i can confirm with 3.14 that this PR breaks transparency of what is set by 
MatePanelApplet .button {}
https://github.com/mate-desktop/mate-themes/blob/gtk3.14/desktop-themes/TraditionalOk/gtk-3.0/mate-applications.css#L133
But the applet itself is transparent.
And with >=3.20 i see always a background on the buttons with or w/o the PR.
So the question is why background-change signal does not have any influence with 3.20/22 ?
I recall that bg handling was changed from 3.18 to 3.20 in mate-panel.

raveit65:
Btw. this turns out that i have to split button styling for applets and wnck-tasklist for 3.20.
Same what i did with other themes.
....sigh...

raveit65:
I really like to understand why background-change signal isn't supported any more with 3.20/22 before i add GTK_VERSION_CHECK to the code?
Is this a result of changes in our code or gtk+ ?

raveit65:
I found more commits like this
https://git.gnome.org/browse/gnome-panel/commit/?id=9c6d738c8ce82772cbd67bbad970001383323ce6
https://git.gnome.org/browse/gnome-panel/commit/?id=4d1d6044f3306b8a297e501f1b62158c7ab2ea07
with a little comment:
```
It is emitted only for out-of-process applets.
```
@monsta 
@lukefromdc 
We can build our applets with in/out-of-process, right?
So it sounds like we need the code.

raveit65:
and here it is removed from panel-applet.c
https://git.gnome.org/browse/gnome-panel/commit/?id=8ab4e78b1d8e7315e700ca67bb4a033159dd6533

lukefromdc:
We can indeed build the panel itself with its included  applets (clock, fish, tray, wncklet) in or out of process, but the additional applets provided by mate-applets are always out of process.

Also, building mate-panel with applets in-process causes nasty redraw issues with GTK 3.22 when a custom background is selected:

https://github.com/mate-desktop/mate-panel/issues/468

lukefromdc:
Also, GTK 3.20 forced a lot of changes to how the panel handles custom backgrounds, which then worked very well with GTK 3.20, only to have custom background handling broken again by GTK 3.22 when in-process applets are built due to the same drawing on native window issue that created the caja GTK 3.22 desktop redraw mess and resulting fixes. Possibly something from https://github.com/mate-desktop/mate-panel/issues/468 might give clues to the panel custom background/in-process mess too?

Issue #174, opened 2015-12-30
Feature: monitor the entire CPU frequency


zaps166:
Now it is impossible to monitor the entire CPU frequency. The applet should have an option to show the highest frequency of all cores/ht.

skoehler:
I don't want to add the applet 8 times just to monitor my quadcore CPU. One applet instance with 8 bars would be enough. Also, I don't want to chose governors per CPU core (or thread, in case of hyperthreading). Setting the governor for all CPU cores/threads at once would be good enough for me. One bar which shows the average over all CPU cores/threads would be good enough for most people, i guess - even though I'd prefer 8 bars. But please don't make me add the cpufreq applet 8 times.

Nivelfacile:
waiting for this, when?

Issue #50, opened 2013-06-23
cpufreq: option to show several cores and processors in one applet instance


Yasen6275:
Is it possible to add "All processors" option in cpufreq applet. In laptop with i3 processor I have to have 4 instances of the applet and click 8 times to change governor or frequency.
##

monsta:
+1 to this, would be nice to have that option.

ulgrimar:
+1 and +2 :-D

PR #234, opened 2017-03-23
cpufreq:set frequency on all cores

lukefromdc:
Partial fix for https://github.com/mate-desktop/mate-applets/issues/174

Cpufreq-applet can still be set to monitor any given core's speed but popup menu now sets governor or userspace speed on all cores together.  Users generally want to change the governor or speed for all the cores at once, but governors presumably could clock up only heavily loaded cores when single or dual threaded applications run so the ability to monitor individual cores is a feature.

### DIFF
--- a/cpufreq/src/cpufreq-selector.c
+++ b/cpufreq/src/cpufreq-selector.c
@@ -18,6 +18,7 @@
  */
 
 #include <config.h>
+#include <sys/sysinfo.h>
 
 #ifdef HAVE_POLKIT
 #include <dbus/dbus-glib.h>
@@ -173,16 +174,20 @@ cpufreq_selector_set_frequency_async (CPUFreqSelector *selector,
 				      guint            cpu,
 				      guint            frequency)
 {
-	SelectorAsyncData *data;
+    guint            cores;
+    cores = get_nprocs() ;
+	for (cpu = 0; cpu < cores; cpu = cpu+1){ 
+	    SelectorAsyncData *data;
 
-	data = g_new0 (SelectorAsyncData, 1);
+	    data = g_new0 (SelectorAsyncData, 1);
 
-	data->selector = selector;
-	data->call = FREQUENCY;
-	data->cpu = cpu;
-	data->frequency = frequency;
+	    data->selector = selector;
+	    data->call = FREQUENCY;
+	    data->cpu = cpu;
+	    data->frequency = frequency;
 
-	selector_set_frequency_async (data);
+	    selector_set_frequency_async (data);
+        }
 }
 
 static void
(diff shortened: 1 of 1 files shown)


lukefromdc:
Rebased after https://github.com/mate-desktop/mate-applets/commit/de39bcb84fd0749264340b9ee663e50034b48e5b

lukefromdc:
Is there any reason this would have to be held for mate-applets 1.19? I am using (and NEEDING) this change now and others have also requested it.  If it can't go into 1.18 we could branch 1.19 early-or just direct users to my branch but then I have to rebase every time any other change comes to mate-applets

raveit65:
> Is there any reason this would have to be held for mate-applets 1.19?

At the moment  we didn't branched any repo to 1.18 and this can be seen as a bugfix, so this might be possible.

monsta:
Are you sure this won't be confusing to some users, given that the applet still shows the info for only one core?

lukefromdc:
Setting the frequency for only one core is what is confusing and I suspect few users are even aware that this is what the applet now does. I thought wrongly that all cores were being controlled (as is the case with Ubuntu's indicator-cpufreq which works this way). This is easily solved, there is no obvious way to deal with the display issue. Fortunately, most of the time the cores seem to change frequencies together (at least on any of my machines) so long as all of them are on the same governor,

The only reason the applet works the way it does is that it was writtten in the age of single-core chips, and the kernel interface controls the cores separately. Theoretically this would let a governor clock up just the core handling a big single-threaded load, but I've never seen any governor on one of my AMD chips actually do that.

If someone comes up with a way to read and display the highest speed used by any core and output that to the applet, that would be a welcome addtion to this PR, but another user has already stated on the orginal issue that what really matters is being able to control all the cores.

lukefromdc:
Possible issue here: I don't have any BSD systems so can't develop directly for them. Turns out get_nprocs is in the GNU C library and I am not sure it is available on BSD systems. Thus this would need to be limited to LInux builds only, not sure how to do that. Looking for solutions

lukefromdc:
I'm not finding ANY platform-independant or even linux+unix(BSD) ways to count CPU cores, needed for this to work and not throw errors by trying to set cores that don't exist

lukefromdc:
Turns out BSD is not a problem, the CpuFreq applet is only built in Linux because it already uses linux-specific code. Apparently none of this can be done in a platform-independent manner. Found this code in configure.ac so this PR won't affect any builds but Linux builds:

```
build_cpufreq_applet=no

if test x$disable_cpufreq = xno; then
   case "${host}" in
      *linux*)
         build_cpufreq_applet=yes
	 ;;
      *)
         AC_MSG_WARN([${host} is not supported by cpufreq applet, not building])
	 build_cpufreq_applet=no
	 ;;
   esac
fi
```

lukefromdc:
Displaying the highest frequency of any core would not be easy. Setting the frequency turns out to be simple, as this uses code that can access all cores, while reading the frequency appears to read from a structure that is set up when the applet is created and only contains data for the particular core it is set to monitor. The monitored core defaults to cpu0 and probably is rarely changed by users. Thus, reading the frequency is not as easily looped to iterate through all cores, which would allow storing and displaying the highest value.

raveit65:
> Are you sure this won't be confusing to some users, given that the applet still shows the info for only one core?

I don't think because using a different governor for a single cpu is a very rare case, and i can't see any use case for what this is good for.
I never thought that the governor was only for on cpu all the years ;-)
I always expected that i had set it for all cpus.
I think this is something which shout be writen in documentation, that's it.
Let's make it not complicated as it is.

monsta:
Ok, I guess you guys know more about this than me :slightly_smiling_face: 
Let's merge it then, and maybe later we'll implement other things that are requested at https://github.com/mate-desktop/mate-applets/issues/50 and https://github.com/mate-desktop/mate-applets/issues/174.

raveit65:
https://github.com/mate-desktop/mate-applets/commit/fbf54b2422909183d4698130408fa160b81e0d70
merged

PR #241, opened 2017-04-11
Two new options for Load graph in Multiload applet

marosg42:
For load average original code calculated maximum y-axis in graph as 5 * (numcpu+1). If you have 4 CPUs it is 25. If your average load (as shown it 'top') is 1.5 (which is something I would like to see clearly in the graph) it is just 1-2 pixels at the bottom. Only if your load goes to 4-5 is is clearly visible in Load graph.
This code introduced two new options - you can select multiplication factor from 1 to 10 (which was 5 in original code). And the second option allows to ignore number of processors. 
It means if I set maximum to 2 and disable multicpu option, I can see average load graph nicely even when my load is below 1. It will max out at 2.
Defaults are set to keep the original behavior.

### DIFF
--- a/multiload/docs/C/index.docbook
+++ b/multiload/docs/C/index.docbook
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<!DOCTYPE article PUBLIC "-//OASIS//DTD DocBook XML V4.1.2//EN" 
+<!DOCTYPE article PUBLIC "-//OASIS//DTD DocBook XML V4.1.2//EN"
 "http://www.oasis-open.org/docbook/xml/4.1.2/docbookx.dtd" [
   <!ENTITY legal SYSTEM "legal.xml">
   <!ENTITY appletversion "1.10.2">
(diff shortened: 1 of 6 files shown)


lukefromdc:
Found the problem: I normally disable recompiling schemas so checkinstall does not pick them up, but just used make-install, so the schema key was missing. recompiled schemas so it works great now.

marosg42:
Luke, status says you requested changes but I don't see anything there.

raveit65:
Just testing....

marosg42:
Any update on this? I s anything needed from my side?

raveit65:
@XRevan86 @monsta 
Can you do a code review please?

raveit65:
Why closing?
We have only 2 ppl who can do a real code review, sometimes it take some time....

marosg42:
I am redoing it, make it simpler. I will also check how code looks in different editor as XRevan86 pointed out.

PR #244, opened 2017-05-09
Autoscale load

marosg42:
Load graph is scaled automatically according to average load value, like old time xsysload program. If Load  grows over 1, there will be a line signalizing load is over 1 and whole graph is scaled accordingly. When Load goes over 2, there will be two lines etc. Similarly when Load goes down,grid lines will be removed.
Color of grid line can be set in properties. Schema was updated to keep the value of grid line color.

Notes for reviewer:
Most of diff is trailing blanks or whitespaces in empty lines which were automatically changed by atom editor.
Real heavy changes are in load-graph.c in function load_graph_draw()

### DIFF
--- a/multiload/docs/C/index.docbook
+++ b/multiload/docs/C/index.docbook
@@ -602,6 +602,10 @@
 		     <entry><para><guilabel>Background</guilabel></para></entry>
 		     <entry><para>that is, no load</para></entry>
 		    </row>
+		    <row valign="top">
+		     <entry><para><guilabel>Grid line</guilabel></para></entry>
+		     <entry><para>Color of grid lines</para></entry>
+		    </row>
 		  </tbody>
 		</tgroup>
 	      </informaltable>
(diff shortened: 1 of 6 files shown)


marosg42:
Load shows Load average (what you see in 'uptime' or  'top' command in first line). It reflect  overall use of system, not just processor, processor can be utilized low, but load can grow, if you have for example a lot of i/o. Processor shows user/sys/wait/steal/idle values, which is what you see in 'top' in line 3 or so.

Old load graph was trying to take into account number of cpus but it does not make sense, so this PR shows load as seen in other tools, ignoring number of processors.

My plan was to play with processor graph later as well, for some people it may be interesting to see per processor use (like in 'top' when you press key '1'.

And maybe nobody besides me is using multiload applet ?

lukefromdc:
Multiload is very handy in a system where no other monitoring tools are installed, and is available  by default in things like Ubuntu MATE live disk/USB installs without requiring network access to install another package. Sure it has plenty of users, hell even IceWM includes a system monitor by default. Without any form of system monitor you have no idea whats going on when the system slows down

raveit65:
Thank you.

Issue #251, opened 2017-06-12
Can't add Disk Mounter applet to a panel


flexiondotorg:
#### Expected behaviour

Right click a panel, find Disk Mounter, click Add and the applet should be added to the panel. 

#### Actual behaviour

Right click a panel, find Disk Mounter, clicking Add appears to do nothing.

#### Steps to reproduce the behaviour

See *Actual behaviour* above.

#### MATE general version

  * MATE Desktop 1.18

#### Package version

  * mate-applets 1.18.1-0ubuntu1

#### Linux Distribution

  * Ubuntu MATE 17.10 daily.

monsta:
I know this issue... it's actually added, but until you'll plug in some drive, it won't show anything.

marosg42:
I decided to give it a try, to change the code. 
What I achieved is - if applet is added and there are no drives to show, it will create an empty unclickable button...
I have all the logic working (when media is inserted, rejected etc), but I would need a help with actual icon. As it is now, icon is empty, it means it is not visible unless you hover over it with a mouse.
This is the actual code from drive-button.c

```
    if (!self->volume && !self->mount) {
	    if (gtk_bin_get_child (GTK_BIN (self)) != NULL)
	      gtk_image_set_from_pixbuf (GTK_IMAGE (gtk_bin_get_child (GTK_BIN (self))), NULL);
	    return FALSE;
    }

```
Could somebody give me a working code which would put there some default icon or maybe DiskMount icon?

marosg42:
Martin, check my PR, it will solve this.
 https://github.com/mate-desktop/mate-applets/pull/257

monsta:
As #257 is merged, it's time to close this.

PR #257, opened 2017-07-05
If the list is empty, it will show media-floppy icon

marosg42:
When the applet is added and there are no external mounts defined in gstab, no media inserted, the applet is not visible becasue it has nothing to show. This change will add a dummy unclickable icon. When some USB is connected or fstab chanegd, it will remove the dummy icon and behave as before. If there is again niothing to show in applet, dummy icon will be shown.
This addresses issue #251.

### DIFF
--- a/drivemount/drive-button.c
+++ b/drivemount/drive-button.c
@@ -121,11 +121,12 @@ drive_button_new (GVolume *volume)
     DriveButton *self;
 
     self = g_object_new (DRIVE_TYPE_BUTTON, NULL);
-    drive_button_set_volume (self, volume);
-    
-    g_signal_connect (gtk_icon_theme_get_default (),
-        "changed", G_CALLBACK (drive_button_theme_change),
-        self);  
+    if (volume != NULL) {
+      drive_button_set_volume (self, volume);
+      g_signal_connect (gtk_icon_theme_get_default (),
+          "changed", G_CALLBACK (drive_button_theme_change),
+          self);
+    }
 
     return (GtkWidget *)self;
 }
@@ -137,10 +138,10 @@ drive_button_new_from_mount (GMount *mount)
 
     self = g_object_new (DRIVE_TYPE_BUTTON, NULL);
     drive_button_set_mount (self, mount);
-    
+
     g_signal_connect (gtk_icon_theme_get_default (),
         "changed", G_CALLBACK (drive_button_theme_change),
-        self);      
+        self);
 
     return (GtkWidget *)self;
 }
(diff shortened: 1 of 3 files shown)


raveit65:
I just tested this in a VM where i use nfs mounts, but i don't see the dummy icon if nothing is mounted.
Did you use a special icon-theme?
I can't test it on bare metal as i have to much mounts in /etc/fstab.

marosg42:
It shows the dummy if nothing is mountable. It means if there are no mounts
defined in fstab and there is no USB disk connected.
So the whole purpose is not to have an empty applet

marosg42:
Look at issue #251, it should explain

raveit65:
I understand what the PR should do, but i don't see the dummy icon if nothing is mounted.

lukefromdc:
I have a main and backup root partition on the encrypted LVM containing my main filesystem (on ALL my machines) so I can't get an empty list to test this with. Will do a test build to look for problems from it, but that's all I can do

marosg42:
Do you have empty applet? Or do you see other drive icons?

raveit65:
disk-mounter
screenshot with a mounted USB-stick an 2 nfs network shares.
Those network shares have no right-click menu.

marosg42:
Could you the the same with current applet? I changed only the code which
deals with nothing to show, old behavior should not be changed

marosg42:
Code uses Monitor to get list of drives and volumes and later it uses it
again to get icons. All I did was to check if both lists are empty and if
so, show dummy icon. If there are special cases which fail also in current
applet, I will try to deal with those as well.

raveit65:
> Could you the the same with current applet? I changed only the code which deals with nothing to show, old behavior should not be changed

Yes it is the same with current applet version.

raveit65:
Not sure if using a network share is a special case, but i know that some people are using ie. a samba share on a multi-iuser server.
Any way, i am fine with as it fixes a non-existent icon if ejecting a usb-stick, if you don't see a solution.

marosg42:
I was testing it in VM with smb shares in fstab. From what you wrote it seems like autofs is a special case. I can have a look at special cases but maybe after this PR is merged to limit number of changes to review  in one PR.

raveit65:
I am fine with that.

raveit65:
merged https://github.com/mate-desktop/mate-applets/commit/82c4e8bd1505d0c2fad1547bbb5d1db97d35b5ed
Thank you

drivemount/drive-button.c
```diff
@@ -880,6 +901,8 @@ drive_button_ensure_popup (DriveButton *self)
 	    g_object_unref (mount);
 	}
     } else {
+      if (!G_IS_MOUNT(self->volume))
```
raveit65: this causes a build warning.
```
drive-button.c: In function 'drive_button_ensure_popup':
drive-button.c:941:7: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
       if (!G_IS_MOUNT(self->volume))
       ^~
drive-button.c:943:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  display_name = g_mount_get_name (self->mount);
  ^~~~~~~~~~~~
```
@marosg42 
Can you please take a look at it?

drivemount/drive-list.c
```diff
@@ -225,33 +238,56 @@ list_buttons (gpointer key, gpointer value, gpointer user_data)
...
 	}
-    }
-
-    return FALSE;
+	else
+	{
+		gtk_widget_show (self->dummy);
+		if (self->orientation == GTK_ORIENTATION_HORIZONTAL)
+		{
+			gtk_container_child_set (GTK_CONTAINER (self), self->dummy,
```
raveit65: and another one.
```
drive-list.c:277:4: warning: 'i' may be used uninitialized in this function [-Wmaybe-uninitialized]
    gtk_container_child_set (GTK_CONTAINER (self), self->dummy,
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          "left-attach", i + 1, "top-attach", 0,
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          "width", 1, "height", 1,
          ~~~~~~~~~~~~~~~~~~~~~~~~
          NULL);
          ~~~~~
```
@marosg42 ^^^^

marosg42:
I will have a look on Monday

marosg42:
https://github.com/mate-desktop/mate-applets/pull/266

PR #264, opened 2017-08-09
avoid deprecated gdk_screen_get_width/height and gdk_screen_width/height

sc0w:
avoid deprecated:

- gdk_screen_get_width/height

- gdk_screen_width/height


raveit65:
thanks

Issue #268, opened 2017-08-24
cpufreq-applet: freq/units labels render partially or not at all


dwtokarski:
#### Expected behaviour
A cpufreq applet placed on the panel should render completely and remain so.

#### Actual behaviour
Almost always, the last column or two of pixels on the units label is missing (e.g. the z in Mhz is missing the right side).

Somewhat less often, more or all the the units label is missing entirely; likewise for the frequency label. On even more rare occasions the applet doesn't render at all, although it's really there as hovering over its known location will pop up the tooltip, left click opens the freq/governors menu, etc.

I have reason to think the total failure to render is a separate issue from the label truncations behavior.

#### Steps to reproduce the behaviour
No special action required. Just log in to a desktop which has one or more cpufreq-applets on a panel. Can also happen after resume from suspend or even any random moment.

#### MATE general version
1.18.0 currently, though this has been happening since at least 1.16.x.

#### Package version
Currently mate-applets-1.18.1-1.fc26.x86_64 but see above.

#### Linux Distribution
Fedora 25, 26

#### Link to downstream report of your Distribution
None yet.

#### Proposed Patch
The patch below addresses the partial labels issue. I've been running this for several days and it seems solid.

The problem seems to have been in cpufreq-applet.c, where the get_max_???_width routines were trying to find the pixel width of various strings by inserting text into a freshly allocated GtkLabel and calling gtk_widget_get_preferred_width() on it, which just returns zero results. Wild guess: the freestanding GtkLabel has no associated font, so the width can't be determined.  At some point someone decided it would be better to get the widths of the applet-> label instead but they're not initialized at first and who knows where the string pointer in that object was pointing. Wildly different values were returned from time to time, and the first time a positive value is acquired it becomes the sticky width for that label in *applet.

The patch gets around this by placing string values in the labels of the applet, then getting the preferred widths. 

```
diff -ur ../mate-applets-1.18.1-ORIG/cpufreq/src/cpufreq-applet.c ./cpufreq/src/cpufreq-applet.c
 --- ../mate-applets-1.18.1-ORIG/cpufreq/src/cpufreq-applet.c   2017-04-06 15:02:40.000000000 -0400
 +++ ./cpufreq/src/cpufreq-applet.c     2017-08-20 16:05:37.056044732 -0400
 @@ -292,7 +292,6 @@
  cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
  {
        GList *available_freqs;
 -      gint   width = 0;
[...]

PR #269, opened 2017-08-29
cpufreq-applet: fixes truncated labels for frequency and units

dwtokarski:
Bug fix for issue #268.

### DIFF
--- a/cpufreq/src/cpufreq-applet.c
+++ b/cpufreq/src/cpufreq-applet.c
@@ -180,8 +180,8 @@ cpufreq_applet_init (CPUFreqApplet *applet)
         applet->popup = NULL;
         applet->monitor = NULL;
 
-        applet->label = gtk_label_new (NULL);
-        applet->unit_label = gtk_label_new (NULL);
+        applet->label = gtk_label_new ("---");
+        applet->unit_label = gtk_label_new ("--");
         applet->icon = gtk_image_new ();
         applet->box = NULL;
 
@@ -292,6 +292,7 @@ static gint
 cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 {
 	GList *available_freqs;
+	const gchar *saved_text;
 
 	if (applet->max_label_width > 0)
 		return applet->max_label_width;
@@ -299,8 +300,17 @@ cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 	if (!CPUFREQ_IS_MONITOR (applet->monitor))
 		return 0;
 
+	saved_text = strdup (gtk_label_get_text (GTK_LABEL (applet->label)));
 	available_freqs = cpufreq_monitor_get_available_frequencies (applet->monitor);
-	while (available_freqs) {
+
+	if (available_freqs == NULL) {
+		/* This only happens when the monitor is not yet ready during
+		 * applet instantiation. In this case the initial label string
+		 * is "---", so just use the width of that.
+		 */
+		gtk_widget_get_preferred_width (applet->label, &applet->max_label_width, NULL);
+
+	} else while (available_freqs) {
 		gint           label_width;
 		const gchar   *text;
 		gchar         *freq_text;
@@ -319,17 +329,25 @@ cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 		available_freqs = g_list_next (available_freqs);
 	}
 
+	gtk_label_set_text (GTK_LABEL (applet->label), saved_text);
+	free (saved_text);
+
 	return applet->max_label_width;
 }
 
 static gint
 cpufreq_applet_get_max_perc_width (CPUFreqApplet *applet)
 {
+	const gchar *saved_text;
+
 	if (applet->max_perc_width > 0)
 		return applet->max_perc_width;
 
+	saved_text = strdup (gtk_label_get_text (GTK_LABEL (applet->label)));
 	gtk_label_set_text (GTK_LABEL (applet->label), "100%");
 	gtk_widget_get_preferred_width (applet->label, &applet->max_perc_width, NULL);
+	gtk_label_set_text (GTK_LABEL (applet->label), saved_text);
+	free (saved_text);
 
 	return applet->max_perc_width;
 }
@@ -338,15 +356,19 @@ static gint
 cpufreq_applet_get_max_unit_width (CPUFreqApplet *applet)
 {
 	gint           mwidth;
+	const gchar   *saved_text;
 
 	if (applet->max_unit_width > 0)
 		return applet->max_unit_width;
 
+	saved_text = strdup (gtk_label_get_text (GTK_LABEL (applet->unit_label)));
 	gtk_label_set_text (GTK_LABEL (applet->unit_label), "GHz");
 	gtk_widget_get_preferred_width (applet->unit_label, &applet->max_unit_width, NULL);
 
 	gtk_label_set_text (GTK_LABEL (applet->unit_label), "MHz");
 	gtk_widget_get_preferred_width (applet->unit_label, &mwidth, NULL);
+	gtk_label_set_text (GTK_LABEL (applet->unit_label), saved_text);
+	free (saved_text);
 
 	applet->max_unit_width = MAX (applet->max_unit_width, mwidth);
 
@@ -358,37 +380,36 @@ cpufreq_applet_get_preferred_width (GtkWidget *widget, gint *minimum_width, gint
 {
 	CPUFreqApplet *applet;
 	gint           labels_width = 0;
-	gint           width;
+	gint           width = 0;
 
 	applet = CPUFREQ_APPLET (widget);
 
-	if (applet->orient == MATE_PANEL_APPLET_ORIENT_LEFT ||
-	    applet->orient == MATE_PANEL_APPLET_ORIENT_RIGHT)
-		return;
+	if (applet->orient != MATE_PANEL_APPLET_ORIENT_LEFT &&
+	    applet->orient != MATE_PANEL_APPLET_ORIENT_RIGHT) {
+	 	if (applet->show_freq) {
+			labels_width += cpufreq_applet_get_max_label_width (applet) + 2;
+		}
 
+		if (applet->show_perc) {
+			labels_width += cpufreq_applet_get_max_perc_width (applet);
+		}
 
- 	if (applet->show_freq) {
-		labels_width += cpufreq_applet_get_max_label_width (applet) + 2;
- 	}
- 
- 	if (applet->show_perc) {
-		labels_width += cpufreq_applet_get_max_perc_width (applet);
- 	}
- 
- 	if (applet->show_unit) {
-		labels_width += cpufreq_applet_get_max_unit_width (applet);
+		if (applet->show_unit) {
+			labels_width += cpufreq_applet_get_max_unit_width (applet);
+		}
+
+		if (applet->show_icon) {
+			gint icon_width;
+
+			gtk_widget_get_preferred_width (applet->icon, &icon_width, NULL);
+			width = (labels_width + icon_width + 2);
+		} else {
+			width = labels_width;
+		}
 	}
 
-	if (applet->show_icon) {
-		gint icon_width;
-
-		gtk_widget_get_preferred_width (applet->icon, &icon_width, NULL);
-		width = (labels_width + icon_width + 2);
-	} else {
-		width = labels_width;
-	}
-
-	*minimum_width = *natural_width = width;
+	if (minimum_width) *minimum_width = width;
+	if (natural_width) *natural_width = width;
 }
 
 static void
@@ -744,8 +765,12 @@ cpufreq_applet_update_visibility (CPUFreqApplet *applet)
         if (changed)
 		g_idle_add ((GSourceFunc)refresh_cb, applet);
 
-        if (need_update)
-                cpufreq_applet_update (applet, applet->monitor);
+	if (applet->need_refresh && applet->show_icon == TRUE)
+		// First rendering w/ icon? Force visibility.
+		cpufreq_applet_pixmap_set_image (applet, 0);
+
+	else if (need_update)
+		cpufreq_applet_update (applet, applet->monitor);
 }
 
 static void
@@ -757,20 +782,6 @@ cpufreq_applet_update (CPUFreqApplet *applet, CPUFreqMonitor *monitor)
         gint         perc;
         guint        cpu;
         const gchar *governor;
-
-        /* Call refresh only the first time.
-         *
-         * Because the label width determinations now modify the actual labels
-         * in *applet to size the max label width in pixels, do this first
-         * before setting the label text below.
-         */
-        if (applet->need_refresh) {
-                gint w;
-
-                cpufreq_applet_get_preferred_width(applet, &w, &w);
-                cpufreq_applet_refresh (applet);
-                applet->need_refresh = FALSE;
-        }
 
         cpu = cpufreq_monitor_get_cpu (monitor);
         freq = cpufreq_monitor_get_frequency (monitor);
@@ -825,6 +836,12 @@ cpufreq_applet_update (CPUFreqApplet *applet, CPUFreqMonitor *monitor)
 		gtk_widget_set_tooltip_text (GTK_WIDGET (applet), text_tip);
 		g_free (text_tip);
 	}
+
+	/* Call refresh only the first time */
+	if (applet->need_refresh) {
+		cpufreq_applet_refresh (applet);
+		applet->need_refresh = FALSE;
+	}
 }
 
 static gint
@@ -869,6 +886,14 @@ cpufreq_applet_refresh (CPUFreqApplet *applet)
 
 	horizontal = (applet->orient == MATE_PANEL_APPLET_ORIENT_UP ||
 		      applet->orient == MATE_PANEL_APPLET_ORIENT_DOWN);
+
+	/* We might be here for any number of reasons, including a font change,
+	 * so the label widths need to be redetermined.
+	 */
+	applet->max_label_width = 0;
+	applet->max_perc_width  = 0;
+	applet->max_unit_width  = 0;
+	cpufreq_applet_get_preferred_width (applet, NULL, NULL);
 
         /* We want a fixed label size, the biggest */
 	if (horizontal)
--- a/cpufreq/src/cpufreq-applet.c
+++ b/cpufreq/src/cpufreq-applet.c
@@ -292,7 +292,6 @@ static gint
 cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 {
 	GList *available_freqs;
-	gint   width = 0;
 
 	if (applet->max_label_width > 0)
 		return applet->max_label_width;
@@ -302,7 +301,6 @@ cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 
 	available_freqs = cpufreq_monitor_get_available_frequencies (applet->monitor);
 	while (available_freqs) {
-		GtkWidget     *label;
 		gint           label_width;
 		const gchar   *text;
 		gchar         *freq_text;
@@ -312,34 +310,26 @@ cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 		freq = atoi (text);
 
 		freq_text = cpufreq_utils_get_frequency_label (freq);
-		label = gtk_label_new (freq_text);
+		gtk_label_set_text (GTK_LABEL (applet->label), freq_text);
 		gtk_widget_get_preferred_width (applet->label, &label_width, NULL);
-		width = MAX (width, label_width); 
+		applet->max_label_width = MAX (applet->max_label_width, label_width);
 
 		g_free (freq_text);
-		gtk_widget_destroy (label);
 
 		available_freqs = g_list_next (available_freqs);
 	}
 
-	applet->max_label_width = width;
-
-	return width;
+	return applet->max_label_width;
 }
 
 static gint
 cpufreq_applet_get_max_perc_width (CPUFreqApplet *applet)
 {
-	GtkWidget      *label;
-	gint            width;
-
 	if (applet->max_perc_width > 0)
 		return applet->max_perc_width;
 
-	label = gtk_label_new ("100%");
-	gtk_widget_get_preferred_width (applet->label, &width,&width);
-	applet->max_perc_width = width +20; /*for some reason width always comes up 2 characters*/
-	gtk_widget_destroy (label);
+	gtk_label_set_text (GTK_LABEL (applet->label), "100%");
+	gtk_widget_get_preferred_width (applet->label, &applet->max_perc_width, NULL);
 
 	return applet->max_perc_width;
 }
@@ -347,20 +337,19 @@ cpufreq_applet_get_max_perc_width (CPUFreqApplet *applet)
 static gint
 cpufreq_applet_get_max_unit_width (CPUFreqApplet *applet)
 {
-	GtkWidget     *label;
-	gint           w1, w2;
+	gint           mwidth;
 
 	if (applet->max_unit_width > 0)
 		return applet->max_unit_width;
 
-	label = gtk_label_new ("GHz");
-	gtk_widget_get_preferred_width (applet->label, &w1, NULL);
+	gtk_label_set_text (GTK_LABEL (applet->unit_label), "GHz");
+	gtk_widget_get_preferred_width (applet->unit_label, &applet->max_unit_width, NULL);
 
-	gtk_label_set_text (GTK_LABEL (label), "MHz");
-	gtk_widget_get_preferred_width (applet->label, &w2, NULL);
+	gtk_label_set_text (GTK_LABEL (applet->unit_label), "MHz");
+	gtk_widget_get_preferred_width (applet->unit_label, &mwidth, NULL);
 
-	gtk_widget_destroy (label);
-	applet->max_unit_width = MAX (w1, w2)-1;
+	applet->max_unit_width = MAX (applet->max_unit_width, mwidth);
+
 	return applet->max_unit_width;
 }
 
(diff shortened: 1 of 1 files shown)


raveit65:
Ok, Menta themes are our default and i like to use it. :-)
So i can test this several days to reproducing the issue.
....with applets for 4 cores.

raveit65:
Well, i can't reproduce the issue.
...now starting to test the fix.

raveit65:
Btw. this looks very bad and should be corrected before merging
```
Author: dwtokarski <dwt@poltec.com>
Date:   Tue Aug 29 17:47:08 2017 -0400

    Bug fix for mate-applets issue #268.
    
    Patch to cpufreq-applet.c fixes truncated labels for frequency and
    units.
```
Better something like this.
```
Author: dwtokarski <dwt@poltec.com>
Date:   Tue Aug 29 17:47:08 2017 -0400

   cpufreq-applet: fixes truncated labels for frequency and units
    
   Bug fix for issue #268.
```
Links belongs always in the body, and the header describes in short what the commit does in code.

lukefromdc:
The short moment of displaying "MHz" may be related to some machines coming up momentarily as "0 MHZ," which was a cause of truncation . 0 is fewer characters than any decimal number of MHz greater than 9.  This may relate to underlying library but presumably not kernel/driver behavior since those should not react to a change in a userspace program. On the other hand, if the kernel driver is momentarily reporting zero each time it is first connected to, that is something we can only work around, not fix here.  If our tools are reporting zero before making that connection at all, that's a problem on this end, and the Ubuntu indicator-cpufreq does not seem to have this issue.

raveit65:
> On the other hand, if the kernel driver is momentarily reporting zero each time it is first connected to, that is something we can only work around, not fix here.

I see that every time if i logged in to session or restart the panel.
So, i am in doubt that the kernel send 'zero' just in all that moments.

dwtokarski:
@raveit65 
  "Btw. this looks very bad and should be corrected before merging"

OK, done. (I finally figured out how to make the changes, slow this week...). Got it right this time, I hope.

I noticed originally when the issue number reference got created as a link, but aside from mild surprise didn't think anything of it. So thanks for the best practices tip!

dwtokarski:
@raveit65 
  "Well, i can't reproduce the issue."

  That's odd, it's pretty easy for me. fwiw, I'm using Fedora 26, sans regular 12pt for the application font, and have a 100 dpi resolution dialed in instead of the default 96. I have four cpufreq applets in my top panel, which is 24 pixels high. 

  I've tried varying those parameters as well as the applet separation, and it doesn't seem to change to frequency of truncation occurrence.

@raveit65, @lukefromdc 
  "If you use 'killall mate-panel' to restart the panel, the applet displays 'MHz'"

  That's happening because "MHz" is the last string written into the units label and it's the only label with a non-zero size at the time of first rendering. There is probably just a garbage pointer in the the freq/perc label, but since producing a non-zero width depends on the availability of a frequencies list (and the monitor has not yet run to provide one) the pixel width of that label remains zero so all you see is "MHz".  It used to be the case that something would render (probably 0, since that's the initial freq value in monitor->priv) simply because 1.18.1 always returns an accidentally-adequate value for the max label width.

In general real data isn't going to be available for display until after the monitor runs once, and that can take up to a second.  Meanwhile, the applet shouldn't just render nothing because it would look like nothing happens at first when adding a new instance to the panel.

I've just pushed up another commit which I think should address all the issues raised here and in the original issue report.

The behavior caused by the one second delay before the monitor runs is interesting. I can't see anything in the monitor code to suggest that an initial call to class->run() couldn't be made in cpufreq_monitor_run() just before setting the timer, so I tried it. Surprisingly it didn't seem to make any difference.

The one outstanding problem which really bugs me is the "invisible rendering" problem, where an applet instance is sometime present in the panel but not visibly rendered. It's there, you can interact with it if you know where it is, you just can't see it. If I have four applets in the top panel and repeatedly kill the panel, then one or more of those applets will render invisibly about one time in ten. I just can't find out why, but suspect the problem is in the panel rather than the applet.

Also, these two commits so far fix the truncation after enlarging the font as observed by @lukefromdc, but only for applets in the TOP panel.  For some reason during a font change, top panel applets get cpufreq_applet_refresh() called on them, but NOT in side or bottom panels. I have no idea...anyone else have a thought?

lukefromdc:
Before applet sucessfully connects and gets the CPU frequency, dashes rather than zero shown, this could prevent a truncated label. 

With this change however, starting the panel with BlueMenta, then changing theme to my UbuntuStudio_Legacy theme (uses a bold font and sets a minumum size (it's larger) to prevent jump from the bold font) will always truncate the last letter in the units label. This does not occur in master. Labels are not truncated when starting/restarting the panel either theme however. 

To test this, just add
```
/*Stop jumping cpufreq labels with bold text*/
#PanelApplet>widget>box>box>label{
	min-width: 35px;
}
```
to any theme, and test vs another theme without it

lukefromdc:
I can also confirm getting the "invisible applet" problem, in a top panel only, and with BlueMenta but not with BlackMATE or UbuntuStudio_Legacy

raveit65:
> I noticed originally when the issue number reference got created as a link, but aside from mild surprise didn't think anything of it. So thanks for the best practices tip!

If you read the commit history 2 years later a link to a bug report in header doesn't say nothing.
So you have to read the report again for understanding the commit ;-)

raveit65:
> With this change however, starting the panel with BlueMenta, then changing theme to my UbuntuStudio_Legacy theme (uses a bold font and sets a minumum size (it's larger) to prevent jump from the bold font) will always truncate the last letter in the units label. This does not occur in master. Labels are not truncated when starting/restarting the panel either theme however. 
>
>To test this, just add
>```
> /*Stop jumping cpufreq labels with bold text*/
> #PanelApplet>widget>box>box>label{
>	min-width: 35px;
>}
>```
> to any theme, and test vs another theme without it

Confirmed with BlueMenta.

raveit65:
> I can also confirm getting the "invisible applet" problem, in a top panel only, and with BlueMenta but not with BlackMATE or UbuntuStudio_Legacy

This i see with BlackMate too if i switch the gtk theme.
But you need only wait some minutes, after a while all cpu-freq-applets are there if you use more than one.
For me it look like the **redraw** after theme change or the **first draw** with panel start, isn't done always for all applets,......only a theory....
Same if you remove another applet, this will trigger a redraw and brings back the invisible applet.

raveit65:
>   "Btw. this looks very bad and should be corrected before merging"

> OK, done. (I finally figured out how to make the changes, slow this week...). Got it right this time, I hope.

>I noticed originally when the issue number reference got created as a link, but aside from mild surprise didn't think anything of it. So thanks for the best practices tip!

Hmm, i do not see any different.
The command for changing existing commits is 'git rebase -i last-good-commit'
I= interactive
This will opens a *Vi* editor.
In your case start with
```
git rebase -i 1def2c13cc2737ebcc082a8a5f469bbde9489fdb
```
Than choose *reword* to change commit header an button.
But at least we will do that before merging;-)

raveit65:
I see build warnings.
```
cpufreq-applet.c: In function 'cpufreq_applet_get_max_label_width':
cpufreq-applet.c:333:8: warning: passing argument 1 of 'free' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  free (saved_text);
        ^~~~~~~~~~
In file included from cpufreq-applet.c:32:0:
/usr/include/stdlib.h:460:13: note: expected 'void *' but argument is of type 'const gchar * {aka const char *}'
 extern void free (void *__ptr) __THROW;
             ^~~~
cpufreq-applet.c: In function 'cpufreq_applet_get_max_perc_width':
cpufreq-applet.c:350:8: warning: passing argument 1 of 'free' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  free (saved_text);
        ^~~~~~~~~~
In file included from cpufreq-applet.c:32:0:
/usr/include/stdlib.h:460:13: note: expected 'void *' but argument is of type 'const gchar * {aka const char *}'
 extern void free (void *__ptr) __THROW;
             ^~~~
cpufreq-applet.c: In function 'cpufreq_applet_get_max_unit_width':
cpufreq-applet.c:371:8: warning: passing argument 1 of 'free' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  free (saved_text);
        ^~~~~~~~~~
In file included from cpufreq-applet.c:32:0:
/usr/include/stdlib.h:460:13: note: expected 'void *' but argument is of type 'const gchar * {aka const char *}'
 extern void free (void *__ptr) __THROW;
             ^~~~


cpufreq-applet.c: In function 'cpufreq_applet_refresh':
cpufreq-applet.c:896:38: warning: passing argument 1 of 'cpufreq_applet_get_preferred_width' from incompatible pointer type [-Wincompatible-pointer-types]
  cpufreq_applet_get_preferred_width (applet, NULL, NULL);
                                      ^~~~~~
cpufreq-applet.c:379:1: note: expected 'GtkWidget * {aka struct _GtkWidget *}' but argument is of type 'CPUFreqApplet * {aka struct _CPUFreqApplet *}'
 cpufreq_applet_get_preferred_width (GtkWidget *widget, gint *minimum_width, gint *natural_width)
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Beside from that it seems to be much better with latest commit.
- displaying only the icon without a value seems to be logic if no data can displayed. no Mhz any more.
- i see always every applet for each cpu if i switch the theme, ok it take some seconds to display all but much better.
```

**Comment by raveit65** [comment]
cpu-freq-applet

...at least it's on my TODO list to  kill deprecated GtkAlignment  ;-)

**Comment by monsta** [comment]
The invisible applet problem isn't new, I get it w/o this PR when I resize the panel (the bottom one in my case), if the CPU frequency doesn't change at that moment.

**Comment by monsta** [comment]
Check https://github.com/mate-desktop/mate-applets/pull/271 - it fixes the invisibility issue for my case, wonder if it'll work for you guys.

**Comment by raveit65** [comment]
The other PR fixes the visibility issue for me.

**Comment by dwtokarski** [comment]
@raveit65 
> Hmm, i do not see any different.
> The command for changing existing commits is 'git rebase -i last-good-commit'
> I= interactive

Ah, that was just me being a git noob. I was looking on github for the editing facility and wound up changing the header for the PR thread.  Should have realized the revision had to originate locally, then push. When I tried that there was a reject because my github branch had supposedly diverged, and I had to do a pull first. Maybe the pull should have preceded the rebase, because that then required another commit message.

Anyway, how does it look now?

**Comment by dwtokarski** [comment]
@monsta 
> Check #271 - it fixes the invisibility issue for my case, wonder if it'll work for you guys.

Yeah, that fixed it. Awesome, thanks!

**Comment by raveit65** [comment]
You can check the result of the commit with adding .patch or .diff to browserlink here.
https://github.com/mate-desktop/mate-applets/pull/269.patch
But we will edit the commit header before we merge it this time, np.
Important is that the new build warnings are fixed.

**Comment by raveit65** [comment]
I could fix the double commits with git rebase, but the build warnings aren't fixed.
```
cpufreq-applet.c: In function 'cpufreq_applet_get_max_label_width':
cpufreq-applet.c:333:8: warning: passing argument 1 of 'free' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  free (saved_text);
        ^~~~~~~~~~
In file included from cpufreq-applet.c:32:0:
/usr/include/stdlib.h:460:13: note: expected 'void *' but argument is of type 'const gchar * {aka const char *}'
 extern void free (void *__ptr) __THROW;
             ^~~~
cpufreq-applet.c: In function 'cpufreq_applet_get_max_perc_width':
cpufreq-applet.c:350:8: warning: passing argument 1 of 'free' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  free (saved_text);
        ^~~~~~~~~~
In file included from cpufreq-applet.c:32:0:
/usr/include/stdlib.h:460:13: note: expected 'void *' but argument is of type 'const gchar * {aka const char *}'
 extern void free (void *__ptr) __THROW;
             ^~~~
cpufreq-applet.c: In function 'cpufreq_applet_get_max_unit_width':
cpufreq-applet.c:371:8: warning: passing argument 1 of 'free' discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
  free (saved_text);
        ^~~~~~~~~~
In file included from cpufreq-applet.c:32:0:
/usr/include/stdlib.h:460:13: note: expected 'void *' but argument is of type 'const gchar * {aka const char *}'
 extern void free (void *__ptr) __THROW;
             ^~~~




cpufreq-applet.c: In function 'cpufreq_applet_refresh':
cpufreq-applet.c:896:38: warning: passing argument 1 of 'cpufreq_applet_get_preferred_width' from incompatible pointer type [-Wincompatible-pointer-types]
  cpufreq_applet_get_preferred_width (applet, NULL, NULL);
                                      ^~~~~~
cpufreq-applet.c:379:1: note: expected 'GtkWidget * {aka struct _GtkWidget *}' but argument is of type 'CPUFreqApplet * {aka struct _CPUFreqApplet *}'
 cpufreq_applet_get_preferred_width (GtkWidget *widget, gint *minimum_width, gint *natural_width)
 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

**Comment by raveit65** [comment]
Warning in *cpufreq-applet.c: In function 'cpufreq_applet_refresh':*
you can fix with
```
cpufreq_applet_get_preferred_width (GTK_WIDGET (applet), NULL, NULL);
```

**Comment by monsta** [comment]
Ok, https://github.com/mate-desktop/mate-applets/pull/271 is merged, one bug less 🙂

**Comment by raveit65** [comment]
Warning in *cpufreq-applet.c: In function 'cpufreq_applet_refresh':*
you can fix with
```
cpufreq_applet_get_preferred_width (GTK_WIDGET (applet), NULL, NULL);
```

**Comment by dwtokarski** [comment]
@raveit65 
> You can check the result of the commit with adding .patch or .diff to link in browser.
> https://github.com/mate-desktop/mate-applets/pull/269.patch
> But we will edit the commit header before we merge it this time, np.
> Important is that the new build warnings are fixed.

Neat trick with the url extension. I see it shows both versions of my Aug 29 and Sept 4 commits. I expected the originals to be supplanted by the new ones during the rebase to fix the commit message. Hope this isn't a problem. I'm tempted to try and remove the old ones, but don't want to break anything else, so I'd best leave that to you.

There's a new commit now for all of the build warnings, including the missing pointer cast you noted later.

**Comment by raveit65** [comment]
> I'm tempted to try and remove the old ones, but don't want to break anything else, so I'd best leave that to you.

Np, i rebased all to  resulting 2 commits local.
Thanks , all warnings are gone and i don't see truncted labels any more with using latest UbuntuStudio_Legacy theme.
Using the style setting in comment https://github.com/mate-desktop/mate-applets/pull/269#issuecomment-327026061
works for me with all themes if there is enough space on the panel to set a **min-width**.
@lukefromdc 
If you can confirm, i am fine with merging the PR.

File: cpufreq/src/cpufreq-applet.c [hunk 1]
```diff
@@ -873,6 +887,14 @@ cpufreq_applet_refresh (CPUFreqApplet *applet)
 	horizontal = (applet->orient == MATE_PANEL_APPLET_ORIENT_UP ||
 		      applet->orient == MATE_PANEL_APPLET_ORIENT_DOWN);
 
+	/* We might be here for any number of reasons, including a font change,
```
monsta: Increasing font size actually still truncates the label... you'll need https://github.com/mate-desktop/mate-applets/pull/274 for that.

raveit65:
@monsta 
I don't see a truncated label if i increase the font-size in our themes. Same with Ubuntu legacy.

lukefromdc:
https://github.com/mate-desktop/mate-applets/pull/274 worked in the test I just did

dwtokarski:
> @monsta
> I don't see a truncated label if i increase the font-size in our themes. Same with Ubuntu legacy.

Actually it does still happen for me, and PR #274 does not help, but I see this only for a cpufreq on the bottom panel. That's what I was talking about in an earlier post:

> Also, these two commits so far fix the truncation after enlarging the font as observed by  
> @lukefromdc, but only for applets in the TOP panel. For some reason during a font change, top 
> panel applets get cpufreq_applet_refresh() called on them, but NOT in side or bottom panels. I have
> no idea...anyone else have a thought?

The absence of a refresh call on a side panel doesn't seem to matter as the label will just expand to fill the width of the panel and doesn't truncate _(why???)_ so long as the panel has room.

When a font change prompts a cpufreq_applet_refresh() call on the top panel cpufreq instance the first ten lines of the backtrace look like this:

    bt
    0  0x00005625e25f21c4 in cpufreq_applet_refresh (applet=0x5625e3c18470) at cpufreq-applet.c:884
    1  0x00007fb88a6e7221 in g_closure_invoke () at /lib64/libgobject-2.0.so.0
    2  0x00007fb88a6f9a5e in signal_emit_unlocked_R () at /lib64/libgobject-2.0.so.0
    3  0x00007fb88a7021a5 in g_signal_emit_valist () at /lib64/libgobject-2.0.so.0
    4  0x00007fb88a702b0f in g_signal_emit () at /lib64/libgobject-2.0.so.0
    5  0x00007fb88c79cc27 in gtk_widget_size_allocate_with_baseline () at /lib64/libgtk-3.so.0
    6  0x00007fb88c7b2dc3 in gtk_window_size_allocate () at /lib64/libgtk-3.so.0
    7  0x00007fb88a6e730d in g_closure_invoke () at /lib64/libgobject-2.0.so.0
    8  0x00007fb88a6f9a5e in signal_emit_unlocked_R () at /lib64/libgobject-2.0.so.0
    9  0x00007fb88a7021a5 in g_signal_emit_valist () at /lib64/libgobject-2.0.so.0
    10 0x00007fb88a702b0f in g_signal_emit () at /lib64/libgobject-2.0.so.0

How that call can happen at all utterly and completely baffles me. The refresh procedure is static and unknown outside its source file. The only places where it's called directly are in that file, and if any of those paths were followed the caller should appear in the backtrace. (I have inspected the entire backtrace and found no hint of this.)

cpufreq_applet_update_visibility() will queue refresh_cb() for later execution by gtk, but refresh_cb() returns false which should dequeue it. What's shown here suggests that gtk somehow remembers the address of cpufreq_applet_refresh() and calls it whenever for reasons unknown.  And only for an instance in the top panel.

With @monsta's patch from PR #274 in place, refresh gets called on side and bottom panel instances too, and if you set a breakpoint in cpufreq_applet_style_updated() you will hit that first. Then single stepping into the refresh procedure causes style_updated to disappear from the backtrace, which then looks like the one above (hmmmm...that's because gcc has replaced what would have been a call with a jump). It looks like compiler optimizations are leaving gdb as confused as me. If I rebuild with -O0 it may be possible to find out where the call is coming from.

Regardless, I still get label truncation on the cpufreq instance on the bottom panel until something else happens to cause the panel to redraw.

monsta:
@raveit65: I actually meant increasing font size in mate-appearance-properties.

@dwtokarski: really weird - I did all the tests with bottom panel and didn't see any truncations. Can you try both https://github.com/mate-desktop/mate-applets/pull/272 and https://github.com/mate-desktop/mate-applets/pull/274?

@lukefromdc: https://github.com/mate-desktop/mate-applets/pull/272 is updated with the fix for themes with bold font.

lukefromdc:
I just combined this PR with https://github.com/mate-desktop/mate-applets/pull/274 and when used together it just plain works, no longer truncating on the change to a theme giving a wider label. If we merge both PRs we are done with this unless someone else finds a problem with that combination

raveit65:
With this PR and https://github.com/mate-desktop/mate-applets/pull/274 i see truncted labels only if i change fontsize with m-a-p and only if applats are at the bottom panel.
With top panel all is good.
We should write a warning ' Don't use it at bottom panel*   :-).
Any ways, the problem isn't exists if you start session with new font-size and i don't think that so much people use those applets at bottom panel.
So, i am fine with merging it.

PS. i do not understand for what 274 is good for. Btw. why is it done?
Does it do the same?
And if yes, why a alternative if we have this one?
This is only confusing for me.....

monsta:
I started digging into problem from https://github.com/mate-desktop/mate-applets/issues/268#issuecomment-327136238 and came up with #272 and later with #274. These two can be used together, but probably not with this one.

This PR seems to be bigger than these ones, so maybe it does something else. I see that @dwtokarski mentions there's still a truncation issue here. So I'm interested in testing #272 + #274, but w/o this one.

dwtokarski:
@monsta 
> @dwtokarski: really weird - I did all the tests with bottom panel and didn't see any truncations. Can you try both #272 and #274?

I think I have an explanation for that, below. I will look at #272 as well.

@raveit65, @monsta,
> PS. i do not understand for what 272 is good for. Btw. why is it done?
> Does it do the same?

> I started digging into problem from #268 (comment) and came up with #272 and later with #274. These two can be used together, but probably not with this one.

PR #272 is an alternate approach to the first commit (b144858) of this PR.  @monsta seeks to use a PangoLayout as a container for the label then get the width of the layout container--vs my approach of setting text in the cpufreq applet label and sizing that.  I don't know enough about Pango to comment on the virtues of #272, but I wonder how it knows what font to assume when performing its rectangle size computation.

#272 does _not_ try to force a from-scratch computation of the label widths when font attributes change.  That he seeks to do in #274, so #272 and #274 together are an alternative to #269.

With regard to truncation in the bottom panel and general weirdness, this has been an instructive exercise.  The tl;dr is that panel height has a lot to do with whether we observe truncation and when it gets fixed, and that #274 has no effect on the observed problem even though it appears to execute as intended.

As before, more experiments are in order. I should have thought before to look and see how mate-sensors-applet works since it doesn't have this problem.

### The gory details...

Building the applet with only this PR and with no code optimizations, gdb was able to show me a properly unwound backtrace.  Setting up a single cpufreq in the top panel with a breakpoint in the refresh procedure, I can hit the breakpoint by using mate-appearance-properties to change the application font point size from 12 to 13. The resulting backtrace is:

    #0  0x0000557b4f30ade8 in cpufreq_applet_refresh (applet=0x557b50b4d6f0) at cpufreq-applet.c:884
    #1  0x0000557b4f30985b in cpufreq_applet_size_allocate (widget=0x557b50b4d6f0, allocation=0x7ffc38421a60) at cpufreq-applet.c:287
    #2  0x00007f698a93f221 in g_closure_invoke () at /lib64/libgobject-2.0.so.0
    #3  0x00007f698a951a5e in signal_emit_unlocked_R () at /lib64/libgobject-2.0.so.0
    #4  0x00007f698a95a1a5 in g_signal_emit_valist () at /lib64/libgobject-2.0.so.0

cpufreq_applet_size_allocate() does indeed call the refresh procedure. 

And yet this still didn't happen for a cpufreq in the bottom panel--until after some flailing around I set the point size to 20, then the bottom panel hit it too.

It looks like something in the gtk libs--or maybe something in the panel--calls gtk_widget_size_allocate_with_baseline() if it thinks something has happened which might cause the applet as currently rendered to not fit the panel properly. That sets off a chain of calls resulting in the backtrace above.

My top and bottom panels all along have been 24 and 32 pixels high, respectively, so there was more space available on the bottom one. This seems to be why a refresh occurs on the top but not the bottom. If both panels are 24 high, both applets get refreshed with just a single point font size change, and there is no truncation on either even for larger fonts. If both are 32 points high neither refresh until a much larger font size change is done, but even so the top panel will do this at 15 points, and the bottom panel not until 20 points. It's odd how the behavior is not the same. Maybe it depends on what else is on the panel. Possibly this whole process is being driven by some other applet and not cpufreq at all.

With PR #274 in place too, and both panels set to 32 pixels high to keep the size_allocate calls from confusing things, any change of the application font will hit a breakpoint in refrsh via cpufreq_applet_style_updated(), as intended. Beyond that the panels' behaviors diverge according to how big the font change is. Consider what happens for a 12pt to 17pt change.

In this case, the top panel will first go through refresh via style_updated. When cpufreq_applet_get_preferred_width() gets called, the widths it determines are still for the smaller 12pt font. When execution is resumed the refresh breakpoint is immediately hit again through the size_allocate path. The applet font, which had previous remained at 12pt has now visibly been redrawn to 17pt and is truncated. The smaller 12pt label widths are still in *applet, but this time with the larger font actually being used, more appropriate larger widths will be computed and the applet redraws with no truncation.

The bottom panel behaves the same way except the refresh via size_allocate does not occur until the font is increased to at least 20pt. Consequently there is a range of font sizes between 15pt and 20pt where truncation does not occur in the top panel but does on the bottom one.

So while PR #274 does force a refresh on font change as intended, the timing is wrong and the desired result of computing label widths appropriate to the new font size isn't achieved. Indeed, with 32 pixel high panels I can't see any difference between having PR #274 in or not.  Whether you see truncation or not really just depends on how tall the panel is plus some other as yet not understood behavior.

monsta:
Ok, so the described details are about this PR, not #272?

dwtokarski:
@monsta: Ok, so the described details are about this PR, not #272?

Yes, that's correct. I was working with #269, alone and in combination with #274.

(btw, sorry to neglect you guys for a few days; some other issues grabbed my attention.)

I said I would also take a look at #272, and just finished doing that, having applied it in combination with #274 and _leaving out_ this PR #269.

And that works. I ran the same set of tests with changing panel heights and font sizes and see no truncation.  Stepping through with the debugger, I see substantially the same behavior as I described above for #269.  The initial refresh call is actually a noop wrt the max label widths in *applet for top and bottom panels, and the get_preferred_width call comes through a different path as before.  This time though, cpufreq_applet_get_max_label_width() uses the new get_max_text_width() with its pango layouts and is able to correctly determine the label widths.  For reasons, may I admit, that I only dimly (think) I understand.  As before, another refresh may occur given a large enough font size increase.

So I would recommend that PRs #272 / #274 be used and this PR #269 be dropped.  Do I need to do anything to make #269 go away, or is it sufficient to simply ignore it?

I have one small change to suggest regarding #272, but I'll go make it in the discussion thread there.

raveit65:
We can close this one later after #272  is merged.
Please don't delete you repo for the moment.

raveit65:
Thanks for all.

PR #272, opened 2017-09-06
cpufreq: fix calculating max width of frequency/unit/percent labels

monsta:
Should work on horizontal panel with GTK+ 3.22.
Vertical panel has some issues, but it seems they were there already.

PangoLayout trick was stolen from GtkScale code:
https://git.gnome.org/browse/gtk+/tree/gtk/gtkscale.c?h=gtk-3-22#n1236

Need to test with GTK+ 3.18 as well...

### DIFF
--- a/cpufreq/src/cpufreq-applet.c
+++ b/cpufreq/src/cpufreq-applet.c
@@ -291,6 +291,24 @@ cpufreq_applet_size_allocate (GtkWidget *widget, GtkAllocation *allocation)
 }
 
 static gint
+get_max_text_width (GtkWidget *widget,
+                    const char *text)
+{
+	PangoContext *context;
+	PangoLayout *layout;
+	PangoRectangle logical_rect;
+
+	context = gtk_widget_get_pango_context (widget);
+	layout = pango_layout_new (context);
+	pango_layout_set_text (layout, text, -1);
+	pango_layout_get_pixel_extents (layout, NULL, &logical_rect);
+
+	g_object_unref (layout);
+
+	return logical_rect.width;
+}
+
+static gint
 cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 {
 	GList *available_freqs;
@@ -304,7 +322,6 @@ cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 
 	available_freqs = cpufreq_monitor_get_available_frequencies (applet->monitor);
 	while (available_freqs) {
-		GtkWidget     *label;
 		gint           label_width;
 		const gchar   *text;
 		gchar         *freq_text;
@@ -314,12 +331,10 @@ cpufreq_applet_get_max_label_width (CPUFreqApplet *applet)
 		freq = atoi (text);
 
 		freq_text = cpufreq_utils_get_frequency_label (freq);
-		label = gtk_label_new (freq_text);
-		gtk_widget_get_preferred_width (applet->label, &label_width, NULL);
+		label_width = get_max_text_width (applet->label, freq_text);
 		width = MAX (width, label_width); 
 
 		g_free (freq_text);
-		gtk_widget_destroy (label);
 
 		available_freqs = g_list_next (available_freqs);
 	}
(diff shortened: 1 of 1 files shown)


raveit65:
Please handle this PR or the other PR for yourself, i am tired to test this applet and i tested too much PRs the last 2 months.

lukefromdc:
truncated_applet

This happens full-time with a theme setting a minumum width or a bold font

monsta:
Added a commit which could fix that. New method is taken from `calculate_minimum_width` function in the clock applet. I don't have any themes with bold font, so I tested it with adding bold font to TraditionalGreen theme:

```css
.mate-panel-menu-bar {
    font-weight: bold;
}
```

lukefromdc:
This latest change STILL truncates the applet if the theme is changed to one giving a wider text. Also, in one instance got a very slight truncation even on the first panel restart though not on the second. Removing the minimum width from my theme did not stop the truncation, though it reduced the extent of it at the price that bold text without it makes the applet jump on frequency change with many fonts.

monsta:
Well, theme update should be handled by https://github.com/mate-desktop/mate-applets/pull/274. Does it work when used together with this one?

monsta:
Ok, I can try that theme, I didn't have it before

monsta:
Ok, I see truncation on GTK+ theme change from TraditionalGreen to UbuntuStudio_Legacy even on git master. Happens with frequency or frequency+units, doesn't happen with percent. So it's an old issue, and I'll merge #274 as it gets harder to test things w/o that fix.

monsta:
Now that #274 is merged and this one is rebased and squashed, I see truncation on units label only. No issues on frequency w/o units or on percent label. Weird.

raveit65:
Why not using the other PR?
In my tests the other one + 274 works perfect in most cases, there is only a minor issue with the applet at bottom panel. But this goes away after a panel restart.
Btw. it's true that all depends on panel size as mentioned here.
https://github.com/mate-desktop/mate-applets/pull/269#issuecomment-329004531

monsta:
@lukefromdc: the truncation mentioned above only happens with font smaller than 10px when your theme is set. Maybe you have any hints?

monsta:
> Why not using the other PR?

It's really bigger than this one, it probably does more stuff than needed.

> there is only a minor issue with the applet at bottom panel

Which one? I think I counted more than one in the last comment there, but maybe I got lost in all that text. If I know what it is, I can check if the issue happens with this one as well.

raveit65:
> Which one? I think I counted more than one in the last comment there, but maybe I got lost in all that text. If I know what it is, I can check if the issue happens with this one as well.

All is described here
https://github.com/mate-desktop/mate-applets/pull/269#issuecomment-328263140

> With this PR and #274 i see truncted labels only if i change fontsize with m-a-p and only if applats are at the bottom panel.
With top panel all is good.
We should write a warning ' Don't use it at bottom panel* :-).
Any ways, the problem isn't exists if you start session with new font-size or restart the panel, and i don't think that so much people use those applets at bottom panel.
So, i am fine with merging it.

monsta:
Ok, I see, truncation again... but I don't see it in that situation with this PR.

monsta:
@lukefromdc: update - I just tested this PR with GTK+ 3.18 and your theme, and there's no any truncations with any font size. My previous tests were with GTK+ 3.22.

monsta:
@lukefromdc: so it was caused by hardcoded minimum width...

```css
/*Stop jumping cpufreq labels with bold text*/
#PanelApplet>widget>box>box>label{
	min-width: 35px;
}
```

It was in `gtk-3.20/panel.css` file. I commented that out, and truncations with font <= 10px were gone.
Which "jumping" was it? Can it be reproduced if you apply this PR?

lukefromdc:
I actually tested that myself, and found that prior to merging https://github.com/mate-desktop/mate-applets/pull/274 I would get truncation of the entire last character with the hardcoded width, and truncation of the last part of the last character without it. A bold font comes up wider (at least with the fonts in Debian) and this also the cause of the applet jumping without the hardcoded minimum width.

I added the hardcoded width to my theme because if I remove it, I get the cpu freq applet jumping (changing width) as the cpu frequency changes. The only other fixes for that in the theme are to use a monospace font or remove the bold font, neither of which is acceptable to me in that theme.

At any rate, I was able in GTK 3.22 to run this PR combined with now-merged https://github.com/mate-desktop/mate-applets/pull/274 and did not get truncation

lukefromdc:
Just retested again, no truncation on theme change to my theme thanks to https://github.com/mate-desktop/mate-applets/pull/274, looks ready to go on this end

dwtokarski:
@lukefromdc 
> Just retested again, no truncation on theme change to my theme thanks to #274, looks ready to go on this end.

Concur. I just finished some testing with #272 / #274, see from this date: 
https://github.com/mate-desktop/mate-applets/pull/269#issuecomment-331038386.

@monsta: There's one small change I'd suggest.  According to the gtk docs, gtk_get_preferred_width() & friends permit either the minimum or natural width pointer args to be NULL depending on what you're interested in getting back.  I've never seen a NULL argument make it all the way down to cpufreq_applet_get_preferred_width() (and I've had a breakpoint on it watching for that), but if it ever happens something interesting could occur and good luck figuring it out.

May I suggest that just before the return, change this

    *minimum_width = *natural_width = width;

to

    if (minimum_width) *minimum_width = width;
    if (natural_width) *natural_width = width;

Oh, well, I just lied...there's really one other thing--there is still the issue that a frequency list doesn't become available until the monitor is running and that can take a second or few.  When adding a cpufreq to a panel it can seem at first as though nothing is going to happen.  You might want to swipe the code I used to set up "---" and "--" as initial label/unit and to forcibly set the icon image upon first instantiation.  I did that originally to avoid showing odd strings before frequencies were available, but it makes for nicer user feedback when an applet instance is added.  You see something right away, and get info when it becomes available.

monsta:
@lukefromdc: but I don't see any jumping on frequency change with this PR and the edited theme. Can you still reproduce it when running this PR?

@dwtokarski: sure, I can add the NULL check later. As for the initial value, maybe we could make the applet load the info faster on start?

lukefromdc:
https://github.com/mate-desktop/mate-applets/pull/272 stops the truncation entirely with my theme, even when switching to it from BlueMenta, as I've said a few comments before.

monsta:
I see - I meant the "jumping" of the label.

monsta:
@raveit65 @lukefromdc 
So what's your resolution on this guys? Are there still some cases where truncation occurs? Or is it ready for merging?

lukefromdc:
I've said several times this is ready to go as far as I can tell. Let's merge this and be done with it

monsta:
Ok, I guess the "jumping label" is gone then (with the modified theme)?

lukefromdc:
Truncation is gone with the unmodified theme as well, works fine with the hardcoded width I use to prevent jumping. Ready to go

raveit65:
@monsta 
no truncation any more , i am waiting to test the PR + suggestions from https://github.com/mate-desktop/mate-applets/pull/272#issuecomment-331047448
Or do you want to add this later?
Any ways, ready to go.

monsta:
Ok, merged. We can address the suggestions later.

PR #275, opened 2017-09-10
GSettings change recognized dynamically

marosg42:
When color is changed in dconf-editor, it is immediately shown in applet, no need to remount.


raveit65:
Thanks.

PR #277, opened 2017-09-22
Add >=cpupower-4.7 compatibility

### LINKED ISSUE #238: doesn't compile with cpupower 4.9.0
On Gentoo, I'm forced to downgrade cpupower 4.5.0 because of mate-applets.
When trying to compile mate-applets with cpupower 4.9.0 installed, I get the following error message:

```
libtool: link: x86_64-pc-linux-gnu-gcc -O2 -pipe -mtune=native -march=haswell -Wl,-O1 -o mate-cpufreq-applet cpufreq-applet.o cpufreq-utils.o cpufreq-prefs.o cpufreq-selector.o cpufreq-popup.o cpufreq-monitor.o cpufreq-monitor-factory.o cpufreq-monitor-procfs.o cpufreq-monitor-sysfs.o cpufreq-monitor-libcpufreq.o cpufreq-monitor-cpuinfo.o  -Wl,--as-needed -lmate-panel-applet-4 -lgtk-3 -lgdk-3 -lpangocairo-1.0 -lpango-1.0 -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lcpupower -ldbus-glib-1 -ldbus-1 -lgobject-2.0 -lglib-2.0
cpufreq-monitor-libcpufreq.o: In function `cpufreq_monitor_libcpufreq_run':
cpufreq-monitor-libcpufreq.c:(.text+0x333): undefined reference to `cpufreq_cpu_exists'
collect2: error: ld returned 1 exit status
```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43470949-doesn-t-compile-with-cpupower-4-9-0?utm_campaign=plugin&utm_content=tracker%2F517519&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F517519&utm_medium=issues&utm_source=github).
</bountysource-plugin>

NP-Hardass:
Drops linux version detection as that relies on linux headers version,
not actual cpupower version (they may be independent, as with Gentoo)

Instead, detect presence of cpupower.h header file.

Starting with cpupower-4.9 in Gentoo, we are installing this header even
though it has not been resolved yet in upstream kernel's cpupower
installer.  Mike Gilbert (floppym@gentoo.org) has told me that he
intends to email the linux-pm mailing list about it in the near future.

This is probably the best solution, assuming that other distros have
this header installed (and eventually that upstream will)

Should resolve #238 and I think it is a better solution than #217

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -374,6 +374,8 @@ fi
 AM_CONDITIONAL(HAVE_LIBCPUFREQ, test x$have_libcpufreq = xyes)
 AC_SUBST(LIBCPUFREQ_LIBS)
 
+AC_CHECK_HEADERS([cpupower.h])
+
 build_cpufreq_applet=no
 
 if test x$disable_cpufreq = xno; then
--- a/cpufreq/src/cpufreq-monitor-libcpufreq.c
+++ b/cpufreq/src/cpufreq-monitor-libcpufreq.c
@@ -19,12 +19,16 @@
  * Authors : Carlos García Campos <carlosgc@gnome.org>
  */
 
+#include <config.h>
+
 #include <glib.h>
 #include <glib/gi18n.h>
 
 #include <stdlib.h>
-#include <linux/version.h>
 #include <cpufreq.h>
+#ifdef HAVE_CPUPOWER_H
+#include <cpupower.h>
+#endif
 #include "cpufreq-monitor-libcpufreq.h"
 #include "cpufreq-utils.h"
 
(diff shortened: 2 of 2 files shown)


monsta:
Well, patch for installing `cpupower.h` had been sent [upstream](https://patchwork.kernel.org/patch/10009599/), but it's not known if it will be accepted. Even if it will be, we still need to build with existing kernels.
Also, as mentioned in https://github.com/mate-desktop/mate-applets/issues/238, there might be other problems with this patch.

PR #278, opened 2017-09-25
Rounding error in draw corrected.

marosg42:
Due to rounding error, graph was not proportionally divided, only observable with highter loadavg (like > 8)


raveit65:
merged https://github.com/mate-desktop/mate-applets/commit/b74e6421f0814e53d52c9634516810d126bfd8be
Thanks

PR #280, opened 2017-10-26
Cpufreq: stop jumping numbers moving units label

lukefromdc:
*Force the frequency label to hold the largest size it ever takes
*Reset label size with any change of theme or font to avoid truncation
*Clean up a nasty mix of spaced and tabs in cpufreq_applet_refresh

Fixes https://github.com/mate-desktop/mate-applets/issues/281

### DIFF
--- a/cpufreq/src/cpufreq-applet.c
+++ b/cpufreq/src/cpufreq-applet.c
@@ -782,6 +782,7 @@ cpufreq_applet_update (CPUFreqApplet *applet, CPUFreqMonitor *monitor)
         gint         freq;
         gint         perc;
         guint        cpu;
+        GtkRequisition  req;
         const gchar *governor;
 
         cpu = cpufreq_monitor_get_cpu (monitor);
@@ -794,6 +795,9 @@ cpufreq_applet_update (CPUFreqApplet *applet, CPUFreqMonitor *monitor)
 
         if (applet->show_freq) {
                 gtk_label_set_text (GTK_LABEL (applet->label), freq_label);
+                /*Hold the largest size set by any jumping text */
+                gtk_widget_get_preferred_size (GTK_WIDGET (applet->label),&req, NULL);
+                gtk_widget_set_size_request (GTK_WIDGET (applet->label),req.width, NULL);
         }
 
         if (applet->show_perc) {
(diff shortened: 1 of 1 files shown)


lukefromdc:
To test:

With master, watch the cpufreq applet closely in the usual icon+number+units mode as the frequency changes. You will see the numbers label push the units label around even though the applet as a whole does not normally jump. When it shrinks again the units label moves back, so it is always in motion with recent kernels. 

Edit your theme to put bold text on the panel (not bold/condensed like in my theme as of now) and this gets worse, but it is present in every theme if you look closely.

Switch to this PR and now the frequency label may push the units once but then holds that width, stabilizing the position of the units label. No problems with truncation should occur on changing themes, fonts, or (in the theme) panel font weight.

raveit65:
I do not see the jumping or moving unit label with using
```
#PanelApplet label {
    font-weight: bold;
}
```
Tested with BlueMenta or BlueSubmarine and using this setting inside gtk3-inspector for freq-applet.
I am using governor ondemand with a range from 2,00 GHz to 2.75 from intel quad-core.
[rave@mother ~]$ uname -r
4.13.9-200.fc26.x86_64
[rave@mother ~]$ rpm -qa mate-applets
mate-applets-1.19.2-1.fc26.x86_64

On my notebook with mate-applets-1.18.1 i see only a truncation for the unit label if i use bold label.

lukefromdc:
The jump is there but it is subtle, I think just 1px at a time with regular fonts and a bit more with bold fonts. To see it you need to put the applet right up against another applet, and then another applet up against it so the whole set will move. Space the applet out from others and you might see an icon move around a bit but nothing more.

A quick test can be done by cycling between Farenheit and Celsius units, which will force a shift from two to three digits in at least one sensor as long as you have anything running below 100C but above 100F. You will see the applet move out to accomdate the larger three digit numbers, then hold that width as they are set back to 2 digits. That's actually a bug in this PR, it should be possible for me to modify the code to reset the minimum width back to zero on a scale change. Meanwhile, it will quickly show the difference.

Otherwise, testing requires a bit of time and fiddling, as you need to run the temperature numbers up and down if (like me) you have no working fan sensors. Farenheit units are best for this again, especially if you have at least one sensor that will alternate between above and below 100F. On a laptop, Celsius units might find a sensor going above 100C, but I'd recommend stopping and cleaning out your fans and cooling fins if you see that.

Running 
`dd if=/dev/urandom of=/dev/null`
is a quick way to load the cpu on one thread and build some heat.  You can alternately run and kill this to force the temperature up and down

monsta:
I never saw this and still can't reproduce.

The `cpufreq_applet_get_max_label_width` function in `cpufreq-applet.c` checks all available frequences and selects the max width from them. How to make this function go wrong?

lukefromdc:
Last comment concerned the wrong PR, the similar one for mate-sensors applet.

On this applet, the jump is NOT of the applet as a whole, but rather of the MHZ/GHZ label as the frequency numbers change. Again it is subtle, just a couple px at most with bold text and maybe 1px with regular text.

lukefromdc:
@raveit65 , I duplicated your exact test, using green-submarine with 

```
#PanelApplet label {
    font-weight: bold;
}
```

added, and did get the jump, again NOT of the whole applet but rather of the GHZ label as the numbers were changed. It was strongest on using the userspace governor to switch between 2.80 GHZ and 4.40 GHZ on my setup. With recent kernels that expose a constantly changing frequency, seeing that label move back and forth even 1 or 2 px can be distracting, even though code in the applet as a whole limits the total applet width so it truncates rather than jumping if it is too wide.

raveit65:
Ok, after testing it again with Blue-Submarine the unit label is simply truncated.
cpu-freq
Now, testing if the Pr fix that hopefully.
I really hate if i got different results between 12hours......

lukefromdc:
The truncation was caused by the unit label being pushed over, and the applet having a fixed width. A test with the git master version of UbuntuStudio_Legacy (with changes to take advantage of this PR) will show the jumping of this label for sure

raveit65:
Weird, after switching to UbuntuStudio_Legacy (latest master) and back to Blue-Submarine with gtk3-inspector the truncation is gone,
Sadly i missed again the jumping with UbuntuStudio_Legacy (latest master) :/
I need to test this without using gtk3-inspector.

lukefromdc:
The applet as a whole DOES NOT jump, only the MHZ/GHZ label jumps, and only by a few px. It's possible that you could have a font in use that does not jump the label because the space taken by the numbers in question combined with your available frequencies always computes to the same width. Keep in mind, I only started getting label jumping this year, and font changes in Debian may have been to blame

raveit65:
> only the MHZ/GHZ

Well, with my proc i see only GHZ, so i can't reproduce it.
I use normal SANS regular fonts, from standard fedora installation.
Any ways, i tested the PR and i see no regressions so far.

raveit65:
merged
https://github.com/mate-desktop/mate-applets/commit/99f9632cc6f08be9770e8f7dc3aae381ddc09e2d
https://github.com/mate-desktop/mate-applets/commit/d0abc327883d8510679ebac9778d9579b11449e9
Thanks

PR #301, opened 2018-01-03
Multiload network scale

marosg42:
When network load gets over 100,000 bytes/s grid line is added and graph scales so that top of graph means 200,000 now. When 200,000 is reached, new line is added and 300,000 is top etc. Like loadavg graph.
Number 100,000 is configurable via schema file. To make it changable via gconf editor, it would need to change from relocatable schema, I will look at it as next stage.

Second commit only reformats the craziest cases of formatting, some parts of code were very hard to read due to bad indentation. I  kept is as two commits so that code changes in first commit are easier to review.


raveit65:
hmm,
i am seeing only white lines with normal panel size and using 100,000 bytes/s grid line.
I am using 50000kb/s max download from my provider.
So the 100,000 bytes/s value is to small.
But with using 165 px panel height it looks like it works.
I think we can't use it without a config option for setting up grid line values.

raveit65:
The scalling works now and i see several grid lines with my bandwidth.
But does changing the color works now?
Or did i missed a hint from you at irc channel?

raveit65:
Really nice idea with the retangle.
And changing the color of the grid line is working again.

raveit65:
...just final testing.
@sc0w @clefebvre @monsta 
Would be nice to git this new feature in 1.20. So please test.

raveit65:
Very nice the options with the thresholds. I think some tooltip infos about the function of thresholds would be helpful.
All this needs translations so we can't push it with 1.20, but maybe with a point release when some translations are done.

marosg42:
I think I will rather update Help screen any objections?

raveit65:
Only a some help about the different thresholds.
I am not really sure if i understand this for myself.

multiload/main.c
```diff
@@ -363,21 +363,31 @@ multiload_create_graphs(MultiloadApplet *ma)
...
+			{ _("Net Load"),     "netload2",  6, GetNet },
 			{ _("Swap Load"),    "swapload", 2, GetSwap },
 			{ _("Load Average"), "loadavg",  3, GetLoadAvg },
 			{ _("Disk Load"),    "diskload", 3, GetDiskLoad }
 		};
 
 	gint speed, size;
-	guint net_granularity;
+	guint net_threshold1;
+  guint net_threshold2;
```
monsta: Oops, indent...

multiload/global.h
```diff
@@ -58,20 +59,20 @@ struct _LoadGraph {
 struct _MultiloadApplet
 {
 	MatePanelApplet *applet;
-
```
monsta: Pushed these cleanups as a separate commit

multiload/linux-proc.c
```diff
@@ -49,47 +49,48 @@ static const unsigned needed_netload_flags =
 void
 GetLoad (int Maximum, int data [5], LoadGraph *g)
 {
-    int usr, nice, sys, iowait, free;
```
monsta: Should we use 2 spaces instead of 4 everywhere...?

multiload/main.c
```diff
@@ -73,17 +73,10 @@ help_cb (GtkAction       *action,
 
  	GError *error = NULL;
 
-#if GTK_CHECK_VERSION (3, 22, 0)
```
monsta: For consistency it would be better to remove only the check blocks that were introduced earlier in this PR.
I'll add the PR which does it for the whole project later.

multiload/load-graph.c
```diff
@@ -399,30 +399,30 @@ load_graph_new (MultiloadApplet *ma, guint n, const gchar *label,
...
+      break;
     }
     default:
-	g_assert_not_reached ();
+      g_assert_not_reached ();
     }
 
     if (g->show_frame)
     {
-	g->frame = gtk_frame_new (NULL);
```
monsta: Corrected some of these tab/space mix cases in master - you'll probably need to rebase

raveit65:
oh, now this branch have conflict :/

raveit65:
@monsta 
Huston we have a problem. 
**This branch cannot be rebased due to conflicts**

marosg42:
Closed as agreed with raveit65, due to conflicts. New clean PR will be opened

Issue #297, opened 2017-12-13
Invest Applet no longer working


mr-mop:
#### Expected behaviour
Invest applet downloads stock quotes

#### Actual behaviour
Invest applet only downloads the following:
"It has come to our attention that this service is being used in violation of the Yahoo Terms of Service. As such, the service is being discontinued. For all future markets and equities data research, please refer to finance.yahoo.com."

#### Steps to reproduce the behaviour
Add some stock symbols to the applet and attempt to refresh

It appears the API that was used has been discontinued.
https://yahoo.uservoice.com/forums/382977-finance/suggestions/32103877-yahoo-stock-quote-api-please-bring-back

#### MATE general version
MATE Desktop Environment 1.8.1

#### Package version
mate-applets (1.8.1+dfsg1-3)

#### Linux Distribution
Debian Jessie

monsta:
If they killed the API, how are we supposed to "fix" it?

flexiondotorg:
Unless there is a drop in replacement API, which reading the thread at Yahoo it seems there will never be, I suggest we drop this applet.

monsta:
I agree. If there's no similar API on any other website, this applet will be simply useless.

mr-mop:
Any use? https://www.alphavantage.co/

monsta:
@flexiondotorg: their "free API key" seems suspicious. Are you familiar with any services which offer API on such terms?

lukefromdc:
I get "No stock quotes are currently available" and this began very recently. 

We need to be very careful about online services that may offer "free" service but be interested in collecting personal information, such as logging all places a unique computer connects from. Fortunately, an applet like this should leak a lot less data than a browser. 

Still, we may want to drop the applet unless we can find another reputable service to connect it to.

monsta:
I'm inclined to drop it (or pass it to someone) also because it's quite specific applet, unlike other ones in the package. I think it's better if it's maintained by someone who knows the problem domain.

alexarnaud:
Le 11/01/2018 à 15:27, monsta a écrit :
> I'm inclined to drop it (or pass it to someone) also because it's quite 
> specific applet, unlike other ones in the package. I think it's better 
> if it's maintained by someone who knows the problem domain.

I agree with you.

Best regards.
-- 
Alex ARNAUD
Visual-Impairment Project Manager
Hypra - "Humanizing technology"

flexiondotorg:
[...]

PR #303, opened 2018-01-13
drop broken invest-applet and all Python dependencies

monsta:
closes https://github.com/mate-desktop/mate-applets/issues/297


lukefromdc:
Question: do we have any community maintained applets for mate-panel that use python and are not in this package? If not we could deprecate introspection support in mate-panel 1.20 and remove it in 1.22, or just drop it entirely as of now. The former option gives a heads-up to anyone with a locally written and maintained applet in python to port it to C or C++.

monsta:
@lukefromdc: yes, for example, mintMenu and its fork, mate-menu, are written in Python and use mate-panel's introspection. The introspection doesn't take a lot of time and effort to maintain actually. It's a generated thing, so it doesn't require so much support as manual Python bindings (which you can see in [mate-menus repo](https://github.com/mate-desktop/mate-menus/blob/master/python/matemenu.c).

lukefromdc:
OK, I just wanted to be sure we were still using introspection for something.

PR #304, opened 2018-01-15
Fix clock applet crashes

raveit65:
Reverting the commits is more a quick and dirty, but i will do the same for fedora to avoid spam in my mailbox from users who files out a rhbz report.
Maybe someone find a real fix for clock-applet crashes with those commits.
So, i will leave that PR open.

Fixes https://github.com/mate-desktop/libmateweather/issues/51


raveit65:
Btw. those commits needs to revert because of reverting commits in https://github.com/mate-desktop/libmateweather/pull/52
as they depends on it.

PR #307, opened 2018-02-04
Scaling network graph using thresholds defined by user

marosg42:
Redo of https://github.com/mate-desktop/mate-applets/pull/301


raveit65:
merged after fixing a few extends.
Thanks for this work and the patient with us.

PR #314, opened 2018-03-20
Fix cpufreq applet spacing on HiDPI displays

vkareh:
The CPU Freq applet overrides the libmate-panel-applet size calculation, so it then needs to account for scaling on its own...


lukefromdc:
Excess space taken by the applet confirmed in master, testing this now...

flexiondotorg:
I'm a little late but just thought I'd add that this resolved the issue I was seeing on Ubuntu MATE 18.04 and no regressions.

Issue #175, opened 2015-12-30
Cpufreq units are frequency value


zaps166:
The cpufreq applet doesn't show units (e.g. GHz) - the units currently are the values. Disabling the units causes a value to disappear.
##

monsta:
Can't reproduce it here - but my applet is built with ```libcpufreq``` support instead of ```libcpupower``` as the latter isn't available in Debian/Ubuntu/Mint. Can you check which one is used in your distro?

zaps166:
I'm using Manjaro, so I think it is the same as in ArchLinux: https://projects.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/mate-applets

monsta:
Ok, it seems to use libcpupower since no build flag is set for that and libcpupower is the default one.

For reference, here are screenshots from my system:

![cpufreq-units-on](https://cloud.githubusercontent.com/assets/5138986/12816307/b5654a9a-cb5c-11e5-9db1-c6bc0d36075d.png)

![cpufreq-units-off](https://cloud.githubusercontent.com/assets/5138986/12816325/c769b4ec-cb5c-11e5-8f31-c4df0e5eff0e.png)

As I understand it, in your system the whole text disappears when you turn the units display off?

zaps166:
Yes:

![01](https://cloud.githubusercontent.com/assets/6239090/12816429/c0b7ff2e-cb4c-11e5-8011-0e5ce0ecc4bc.png)
![02](https://cloud.githubusercontent.com/assets/6239090/12816428/c08f592a-cb4c-11e5-86e1-fcc771bd57a7.png)

monsta:
Really weird...

zaps166:
Yes..., but the real problem is here: https://github.com/mate-desktop/mate-applets/issues/173 which makes the applet unusable.

//Anybody knows why the Clock Applet which updates seconds (and Temperature Applet) lock for a while full screen OpenGL application with NVIDIA proprietary + AllowFlipping (NetworkMonitor also does an update very often, but it doesn't interfere with OpenGL...)?

zaps166:
The same reason as in https://github.com/mate-desktop/mate-applets/issues/173 - "intel_pstate" is causing this issue.

gav-:
[...]

PR #320, opened 2018-06-08
Ensure CPU frequency label is always displayed if selected

lukefromdc:
Fix https://github.com/mate-desktop/mate-applets/issues/175 by removing old GTK 2 leftovers that never worked right and using  gtk_widget_get_preferred_size and gtk_widget_get_preferred_width to so the same job.  Same coding used to set frequency label, unit label, and percentage label widths. This also finally gets rid of the maximum applet size that under so many conditions clipped labels and apparently was responsible for #175

### DIFF
--- a/cpufreq/src/cpufreq-applet.c
+++ b/cpufreq/src/cpufreq-applet.c
@@ -64,10 +64,6 @@ struct _CPUFreqApplet {
         GtkWidget        *container;
         cairo_surface_t  *surfaces[5];
 
-	gint              max_label_width;
-	gint              max_perc_width;
-	gint              max_unit_width;
-
 	gboolean          need_refresh;
 
         CPUFreqPrefs     *prefs;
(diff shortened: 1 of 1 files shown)


lukefromdc:
Also tested in a vertical panel, giving these results:

Vertical panels narrower than 30px with 10pt bold text do not have enough space to show the labels, they will appear but be cut off.

vertical panels 30-65px with 10pt bold text will show the applet cleanly with labels below oneanother

vertical panels from 65 to about 94px with 10pt bold text will show all of the applet on one level but will clip the units label to fit

vertical panels wider than 94px with 10pt bold fonts will show the entire applet on one level with no cutting of lables

lukefromdc:
I will have to look at that later as I am about to head out for today, but that looks like I may need to pre-initialize these variables in some way, so they can never be NULL

lukefromdc:
OK, I just found I do NOT get the errors on the desktop, which is AMD and did not have the original bug. Think I see what is going on: With some Intel machines the frequency label does not get any non-zero width until the driver sends input, and that must be occuring late. Thus the errors may be something I will only see in the laptop. To stop them will require initializing the variables to something other than zero before calling  if the warnings are coming from gtk_widget_get_preferred_width, or maybe switching to gtk_widget_get_preferred_size and pulling the data from the allocation.width in the returned value?  Also  I know that minimum values set in CSS will now work, so that might be something that can be quickly tried in the laptop and could be used if effective. Will try that tonight or tomorrow, very busy weekend.

lukefromdc:
Just thought of a better approach: I think at some point before we had tried initializing the applet label as "--" when there was no input. I will see if that was lost, because that would ensure the label width can never be NULL.

lukefromdc:
I just fixed an error where the & had been left out of 
gtk_widget_get_preferred_width(GTK_WIDGET(applet->icon), &pixmap_size, NULL);
see if that fixes the runtime warning

lukefromdc:
OK, looking at the code, the frequency label really ought to come back zero if nothing is found, as it did if I sent "NULL" or "0" to cpufreq_utils_get_frequency_label(freq) . In turn, freq is a gint that comes from cpufreq_monitor_get_frequency (monitor); and that returns a gint as well, copied from monitor->priv->cur_freq in struct struct _CPUFreqMonitorPrivate in the cpufreq-monitor.c file. Only thing is, if I go through every instance of cur_freq in that file, I don't see where it is reliable initialized, and I wonder if this is causing odd behavior in only certain machines.

lukefromdc:
No sign of the warnings on my laptop with any version of this, nor my desktop

lukefromdc:
OK, I think I have traced the warning @vkareh traced to the typo I fixed in 
https://github.com/mate-desktop/mate-applets/pull/320/commits/a3633ef1f6834f3f93c43bf66e5176a3bde10230
That typo meant that instead of submitting a variable to store the return value from gtk_widget_get_preferred_width , I was sending pixmap_size as an input. The code in GTK itself for gtk_widget_get_preferred_width is this:

```
void
gtk_widget_get_preferred_width (GtkWidget *widget,
                                gint      *minimum_width,
                                gint      *natural_width)
{
  g_return_if_fail (GTK_IS_WIDGET (widget));
  g_return_if_fail (minimum_width != NULL || natural_width != NULL);

  gtk_widget_compute_size_for_orientation (widget,
                                           GTK_ORIENTATION_HORIZONTAL,
                                           -1,
                                           minimum_width,
                                           natural_width,
                                           NULL, NULL);
}

```
So with the typo, this was getting an imput where it was supposed to get a place to send an output, and if anything other than NULL came up, the  g_return_if_fail (minimum_width != NULL || natural_width != NULL) is invoked, and that was exactly the warning reported.  Thus it should now stay gone.

lukefromdc:
Rebased and squashed the build warning fix

lukefromdc:
I propose this be cherrypicked for 1.20, as we've got folks who need it as reported by https://github.com/mate-desktop/mate-applets/issues/175 and it will be a while before 1.22 is out and distros are shipping it.

PR #327, opened 2018-07-10
cpufreq: stop segfault with gtk3.23

lukefromdc:
Also stop any other segfaults caused by failure of deprecated gtk_action_group_get_action to return the name of the active item. With GTK 3.23.0 cpufreq_popup_menu_set_active sends invalid name information for userspace governors.

### DIFF
--- a/cpufreq/src/cpufreq-popup.c
+++ b/cpufreq/src/cpufreq-popup.c
@@ -437,7 +437,16 @@ cpufreq_popup_menu_set_active_action (CPUFreqPopup   *popup,
 
 	g_snprintf (name, sizeof (name), "%s%s", prefix, item);
 	action = gtk_action_group_get_action (action_group, name);
-	
+
+	/* gtk_action_group_get_action can return NULL with frequencies (userspace governor)
+	 * when the CPU does not actually stay locked to that exact frequency but rather to a
+	 * frequency range. Since cpufreq_monitor_get_frequency gets the realtime frequency, this
+	 * may not match any named frequency and then returns NULL. Return when this happens to
+	 * avoid segfaults
+	 */
+	if (action == NULL)
+		return;
+
 	g_signal_handlers_block_by_func (action,
 					 cpufreq_popup_frequencies_menu_activate,
 					 popup);


lukefromdc:
Checked with both GTK 3.22.30 and 3.23.0. This simply adds a check to stop segfaults if for any reason gtk_action_group_get_action fails. Also removes a dangling tab, thus the seeming replacement of an empty line. Since GtkActionGroup is getting buggy here and there are no icons in this particular menu, we should think about replacing it soon in this applet.

lukefromdc:
Running cpufreq from master with GTK 3.23.0 resulted in this relevent portion of the backtrace collected from systemd-coredump after the applet crashed, pointed right to the spot where a check was needed:

```
       Message: Process 21802 (mate-cpufreq-ap) of user 1000 dumped core.
                
                Stack trace of thread 21802:
                #0  0x00007fc3422f19a0 gtk_toggle_action_set_active (libgtk-3.so.0)
                #1  0x00005568b9bbb607 cpufreq_popup_menu_set_active_action (mate-cpufreq-applet)
                #2  0x00005568b9bbbdb6 cpufreq_popup_menu_set_active (mate-cpufreq-applet)
                #3  0x00005568b9bb9229 cpufreq_applet_menu_popup (mate-cpufreq-applet)
                #4  0x00005568b9bb9589 cpufreq_applet_menu_popup (mate-cpufreq-applet)
                #5  0x00007fc3425578f9 n/a (libgtk-3.so.0)
                #6  0x00007fc3406efc06 n/a (libgobject-2.0.so.0)
                #7  0x00007fc34070aff4 g_signal_emit_valist (libgobject-2.0.so.0)
                #8  0x00007fc34070bce2 g_signal_emit (libgobject-2.0.so.0)
                #9  0x00007fc342514e24 n/a (libgtk-3.so.0)
                #10 0x00007fc34241366e n/a (libgtk-3.so.0)
                #11 0x00007fc342415422 gtk_main_do_event (libgtk-3.so.0)
                #12 0x00007fc34218aaf5 n/a (libgdk-3.so.0)
                #13 0x00007fc3421b10a2 n/a (libgdk-3.so.0)
                #14 0x00007fc340413627 g_main_context_dispatch (libglib-2.0.so.0)
                #15 0x00007fc340413838 n/a (libglib-2.0.so.0)
                #16 0x00007fc340413b42 g_main_loop_run (libglib-2.0.so.0)
                #17 0x00007fc3424147b5 gtk_main (libgtk-3.so.0)
                #18 0x00007fc342869609 n/a (libmate-panel-applet-4.so.1)
                #19 0x00005568b9bb7f38 main (mate-cpufreq-applet)
                #20 0x00007fc340202b17 __libc_start_main (libc.so.6)
                #21 0x00005568b9bb7f7a _start (mate-cpufreq-applet)
      
```

cpufreq/src/cpufreq-popup.c
```diff
@@ -437,7 +437,11 @@ cpufreq_popup_menu_set_active_action (CPUFreqPopup   *popup,
 
 	g_snprintf (name, sizeof (name), "%s%s", prefix, item);
 	action = gtk_action_group_get_action (action_group, name);
-	
+
+	/*Do not segfault if gtk_action_group_get_action fails*/
+	if(!action || action == NULL)
```
muktupavels: 1) Is not `!action` and `action == NULL` same?
2) That function is documented to return `NULL` if action does not exist for requested name.

Quickly looking it seems that for userspace governors actions is not created... So I am quite surprised that it crash only with GTK 3.23.

muktupavels: Spoke too fast, seems that that code assumes that action will always exist. I would add add two debug printf... One in `cpufreq_popup_menu_add_action` to see all action names that are created and other in `cpufreq_popup_menu_set_active_action` to see requested action name.

lukefromdc:
By playing with g_warning statements to record data in syslog, I just found that the frequency numbers being sent to cpufreq_popup_menu_set_active_action by cpufreq_popup_menu_set_active do NOT match the list of frequency numbers initially loaded by cpufreq_popup_menu_add_action. I don't know yet if this is a formatting issue, or underlying libraries returning the exact CPU frequency at the instant or what, but this now confirmed to be causing `action = gtk_action_group_get_action (action_group, name);` to return NULL, as there is no way it could return a match when it is being sent the wrong frequency name.

muktupavels:
Why did not paste here for what frequencies actions are created and example of action name that is not found?

lukefromdc:
OK, it looks like this is caused by the behavior of modern kernels with modern CPU's in the userspace governor. With libcpupower, a recent Linux kernel, and my AMD FX-8120, the CPU does not actually sit on the named frequency of a userspace governor unless fully loaded. Normally it will dither around a narrow range of frequencies near that governor frequency. Thus,

`freq = cpufreq_monitor_get_frequency (popup->priv->monitor);`

now gives not the nominal frequency for which the selected frequency within the userspace governor is named, but the actual instantanious CPU frequency. This does not match the named frequency, so` action = gtk_action_group_get_action (action_group, name)` returns NULL and with master and GTK 3.23 the result is a segfault. 

The practical result of this is that with master and GTk 3.22 or this PR and GTk 3.23, if the userspace governor is in use and the panel is restarted, the frequency gets reset to a default which at least in my case is the highest value.

I don't see an easy way to fix that, as in cpufreq-monitor.c the structure containing all the information about governors and frequences is this:
```

struct _CPUFreqMonitorPrivate {
        guint    cpu;
        gboolean online;
        gint     cur_freq;
        gint     max_freq;
        gchar   *governor;
        GList   *available_freqs;
        GList   *available_govs;
        guint    timeout_handler;

        gboolean changed;
};
```
The available_freqs value with today's kernels and CPU's comes up as a list of theoretical values, but cur_freq is coming up as what appears to be the instantanious CPU frequency. I will document this at the NULL check in cpufreq-popup.c

lukefromdc:
Here are some examples from two different runs, logging code may not have been the same. Had to go back into my system log to fetch them:
```

Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 4300000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 4300000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 2800000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 2800000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 2300000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 2300000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 1900000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 1900000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 1400000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: 1400000
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: conservative
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: conservative
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: ondemand
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: ondemand
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: performance
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: performance
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: powersave
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: powersave
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: schedutil
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: schedutil
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: Governorondemand
Jul 10 15:47:40 ubuntu mate-cpufreq-ap[31125]: Governorondemand

Jul 10 15:42:26 ubuntu mate-cpufreq-ap[22118]: Frequency3327686
Jul 10 15:42:26 ubuntu mate-cpufreq-ap[22118]: Frequency3327686
Jul 10 15:42:27 ubuntu mate-cpufreq-ap[22118]: Frequency3621669
Jul 10 15:42:27 ubuntu mate-cpufreq-ap[22118]: Frequency3621669
Jul 10 15:42:29 ubuntu mate-cpufreq-ap[22118]: Frequency2755541
Jul 10 15:42:29 ubuntu mate-cpufreq-ap[22118]: Frequency2755541
Jul 10 15:42:30 ubuntu mate-cpufreq-ap[22118]: Frequency2067278
Jul 10 15:42:30 ubuntu mate-cpufreq-ap[22118]: Frequency2067278
Jul 10 15:42:32 ubuntu mate-cpufreq-ap[22118]: Frequency1878587
Jul 10 15:42:32 ubuntu mate-cpufreq-ap[22118]: Frequency1878587
Jul 10 15:42:33 ubuntu mate-cpufreq-ap[22118]: Frequency1461083
Jul 10 15:42:33 ubuntu mate-cpufreq-ap[22118]: Frequency1461083
Jul 10 15:42:34 ubuntu mate-cpufreq-ap[22118]: Governorondemand
Jul 10 15:42:34 ubuntu mate-cpufreq-ap[22118]: Governorondemand
Jul 10 15:42:35 ubuntu mate-cpufreq-ap[22118]: Governorperformance
Jul 10 15:42:35 ubuntu mate-cpufreq-ap[22118]: Governorperformance
Jul 10 15:42:36 ubuntu mate-cpufreq-ap[22118]: Governorpowersave
Jul 10 15:42:36 ubuntu mate-cpufreq-ap[22118]: Governorpowersave
Jul 10 15:42:37 ubuntu mate-cpufreq-ap[22118]: Governorschedutil
Jul 10 15:42:37 ubuntu mate-cpufreq-ap[22118]: Governorschedutil
Jul 10 15:42:38 ubuntu mate-cpufreq-ap[22118]: Governorschedutil
Jul 10 15:42:38 ubuntu mate-cpufreq-ap[22118]: Governorschedutil
Jul 10 15:42:40 ubuntu mate-cpufreq-ap[22118]: Frequency4117335
Jul 10 15:42:40 ubuntu mate-cpufreq-ap[22118]: Frequency4117335
Jul 10 15:42:41 ubuntu mate-cpufreq-ap[22118]: Frequency2354381
Jul 10 15:42:41 ubuntu mate-cpufreq-ap[22118]: Frequency2354381
Jul 10 15:42:43 ubuntu mate-cpufreq-ap[22118]: Frequency2234029
Jul 10 15:42:43 ubuntu mate-cpufreq-ap[22118]: Frequency2234029
Jul 10 15:42:45 ubuntu mate-cpufreq-ap[22118]: Frequency1483505
Jul 10 15:42:45 ubuntu mate-cpufreq-ap[22118]: Frequency1483505
```

lukefromdc:
The dual check `if(!action || action == NULL)` was retained mostly as a check against future bugs, since the current problem is getting NULL back. I see that kind of check used a lot elsewhere in MATE.
As for the space, I had no idea that was an issue.

muktupavels:
Space is for coding style, stick to same style... Will try differently - explain meaning  of `!action` and `action == NULL`. Or why you need both?

lukefromdc:
Not sure I can, this is just what I've seen before. My guess is this is used elewhere because the variable could be empty, or could not exist at all

muktupavels:
Then why are you copy-pasting code that you don't understand? They are same, they both check if `action` is `NULL`, so your if reads like:

> if action is null or if action is null then return

lukefromdc:
Because if don't so something, I cannot use the applet due to the segfaults.

If the checks are the same one is enough, but as I said this same type of check is used all over MATE so I figured they caught related but not exactly the same situations.

Question: Can there  ever be a circumstance where a variable can pass the  !(variable=NULL) check but does not exist at all and fails a !variable check? If not, this needs to be stripped out not only of this PR but of all of MATE

muktupavels:
Did not understand question... Show at least some places where same thing is used!?

lukefromdc:
A manual search of the codebase would be difficult given the number of possible variable names. Not worth the trouble, this will come up when it is seen by chance.

The question was this: Is a variable set to NULL really the same as the variable not existing? I can try the check `if(!action)` by itself and see what happens

muktupavels:
How can variable not exist?

lukefromdc:
Using `if(!action)` works by itself, so does `if(action == NULL)` and the latter is a more human readable match for the comment above it.  

The variable could not exist if it was freed (like the unref problems with the caja pathbar) but after it is first declared the only thing that happens is calling (deprecated and no doubt unmaintained) gtk_action_group_get_action  to set its value, and that returns either the action if a match is found or else NULL . Since it cannot free the variable and nothing else touches it, it can't be freed and can only be a valid action to send to gtk_toggle_action_set_active or else NULL .  

Interestingly, at least this particular call of gtk_toggle_action_set_active() segfaults if the GtkToggleAction value is NULL only with GTK 3.23 and simply fails without segfaulting with GTK 3.22.30 and earlier. It can't be getting a valid action with the earlier GTK versions, as that does not change the kernel, libcpupower, or hardware versions and thus cannot make the reported  CPU frequency match the frequency name in the list of actions.

This sort of thing is related to but not the same as https://github.com/mate-desktop/mate-applets/issues/329 , where the governor is not saved and cannot be saved acros reboots. I pointed out there that even if such code was added, userspace governor settings would remain unsavable due to this frequency mismatch issue that forces gtk_action_group_get_action  to return NULL.

raveit65:
> I have no idea what's holdin that up. GTK is not that hard to build locally, I've done it for years
EDIT: and I've had no trouble with the gtk3.23 builds beyond chasing the usual runtime bugs in programs using it

I tried to build a gtk3-3.23.0 RPM for fedora rawhide, but no luck here.
```
  CC       libgtk_3_la-gtkgesturelongpress.lo
gtkfontchooserwidget.c:63:10: fatal error: language-names.h: No such file or directory
 #include "language-names.h"
          ^~~~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [Makefile:6218: libgtk_3_la-gtkfontchooserwidget.lo] Error 1
make[3]: *** Waiting for unfinished jobs....
```
And i got same build error with fedora 28
Looks like 3.23.0 is far away from an alpha or beta state, or is it in debian unstable?

Well, i can test your PR with gtk+-3.22.30, but i will be unable to reproduce your segfault.

muktupavels:
Any chance that with gtk 3.22 it does not crash, but instead it logs some critical warnings?

`gtk_action_group_get_action` will return `NULL` if there is no action for requested name and that part should not change between gtk versions... Can you verify that with gtk 3.22.x it still returns `NULL` when action does not exist?

Any chance that there is build differences between gtk 3.22 and gtk 3.23? For example do you have `G_DISABLE_CHECKS` defined in 3.23 build?

> g_return_if_fail (GTK_IS_TOGGLE_ACTION (action));

`GTK_IS_TOGGLE_ACTION` is `NULL` safe so if you pass `NULL` action it just returns (and logs critical). If `G_DISABLE_CHECKS` is defined then this check is not performed and `action->private_data->active` obviously will cause segfault.

raveit65:
> Any chance that with gtk 3.22 it does not crash, but instead it logs some critical warnings?

Topic from the PR if for 3.23.0, and No, i don't see that warnings and segfault with stable 3.22.30

muktupavels:
There will be no segfault, I think there is just critical warning. In any case checking for `NULL` there seems like valid fix, but that has nothing to do with gtk 3.23...

raveit65:
Again, i don't see any critical warning in logs with gtk+-3.22.30

muktupavels:
Your configuration and/or cpu might be different from his and you might not have problems. Lets wait until @lukefromdc can test that.

raveit65:
I am using `Governor ondemand` with one single applet monitoring CPU-0 with a AMD Ryzen 5 1400 4/8 core/virtual cores.
A simple standard config.
I know that Luke use also an AMD 8 core.
And i don't see any comment from Luke here, where he mention that warnings and segfault are with gtk+-3.22.30, only 3.23 is mentioned here https://github.com/mate-desktop/mate-applets/pull/327#issuecomment-403666855
So i still wondering why you saying this have to be in 3.22 too without practical testing :)

muktupavels:
Because there has not been any changes in code that could have changed behavior of `gtk_action_group_get_action`:
https://gitlab.gnome.org/GNOME/gtk/commits/gtk-3-24/gtk/deprecated

So it can return `NULL` with gtk 3.23, it can with 3.22 and same was true even with GTK+ 2:
https://developer.gnome.org/gtk2/stable/GtkActionGroup.html#gtk-action-group-get-action

And I doubt that cpu reports different frequencies simply by changed gtk+ version, do you think otherwise? So that function is returning and will return `NULL` if action has not been created for requested frequency no matter what gtk version you are using.

GLib and GTK has checks that verifies that you are passing correct data to functions. If that is not case then function typically generates critical warning and returns. So you get critical warning, but you don't crash...

Luke already confirmed that frequencies does not match! And that is not affected by GTK+ or I would love to learn how GTK does that if that is the case... That means that same problem is also with GTK 3.22. Now why it does not crash with GTK 3.22, but crash GTK 3.23?

One option is that build options are different and `G_DISABLE_CHECKS` is defined making `g_return_*` function to do nothing... So no more warning, instead you get crash. Maybe build differences are in glib build.

Luke easily can verify that by adding this and testing with gtk 3.22:
```
if (action == NULL)
    g_warning ("action == NULL");
```

raveit65:
All i see is that that gtk+-3.23.0 is broken because it doesn't build, so i assume that this pre-alpha version has more issues.
I suggest to wait if gtk+-3.23.x is somewhere build in distros, to be sure that all isn't caused by gtk+ itself.

muktupavels:
https://gitlab.gnome.org/GNOME/gtk/commit/e8d38994611df4b44c7418279833a8f5ab0d1daa and https://gitlab.gnome.org/GNOME/gtk/commit/b9b62fca0c4c7cc1f498b712b21ac70180b1eaf9

So yes, it seems that building from tarball will fail, but these are already fixed and again is unrelated to this problem.

raveit65:
sigh, complete useless without language-names.h in tarball.
I am out at this point.

raveit65:
@lukefromdc 
Feel free to merge it as it is a small code change.
Personal i would wait with merging fixes for 3.23/24 when another person is able to reproduce it.
But....your decision.

lukefromdc:
I will merge this as the existing call is incorrect regardless of GTK version, and also someone with different build options could potentialy get problems according to @muktupavels. All my GTK builds use the same build options so a change there did not cause my problems. Merging this now makes my maintainance a bit easier, and the check catches something that has been wrong since it began in any GTK version and does not come from GTK(the frequency mismatch).

I've already verified with test g_warning that I get  NULL with the frequency mismatch. That is not where the segfault occurs, rather I proved by commenting out` gtk_toggle_action_set_active` that  the segfault occurs when that is called with (NULL, TRUE) instead of (<someaction>, TRUE). I didn't see anything in documentation calling that function null safe. The check just makes the function return before that can be reached if gtk_action_group_get_action returns NULL, as it does in the userspace governor when the CPU frequency does not match any listed "available frequency." 

In no case does gtk_toggle_action_set_active actually WORK with the userspace governor and the frequency mismatch, I tested that in master and with this PR with GTK 3.22.30 to ensure no regression resulted from this. In both cases, if a userspace governor is set and the panel is restarted, the userspace governor defaults to its highest setting, as the user's last frequench choice  did not put the CPU on a frequency matching any of the other available frequencies. With GTK 3.22.30 on my system, that fails silently and the governor defaults to its highest setting. With GTK 3.23.0 and master the applet segfaults when clicked if the userspace governor is active. With GTK 3.23.0 and this PR the behavior is an exact match for master with GTK 3.22.30.

https://developer.gnome.org/gtk3/stable/GtkToggleAction.html#gtk-toggle-action-set-active

muktupavels:
@lukefromdc does `gtk_action_group_get_action` return `NULL` with GTK 3.22? It seems that you tested it only with 3.23... Or I missed something?

Where did I say that `gtk_toggle_action_set_active` is `NULL` safe? I told that only `GTK_IS_TOGGLE_ACTION` is `NULL` safe and it returns `FALSE` in that case...

lukefromdc:
You didn't say gtk_toggle_action_set_active , rather you were discussing gtk_action_group_get_action, which is not the source of the segfault. I will run a final recheck for the action value with GTK 3.22.30 and the added logging function, and will merge on confirmation that I am getting NULL with both GTK versions.

lukefromdc:
OK, here's what I got with some logging test code and this PR, behavior identical with GTK 3.22.30 and the July 2 pull from the GTK 3.24 branch (versioned 3.23.0 in configure.ac):

Test code used:

```
    g_warning("%s", name);
	if (action == NULL){
        g_warning ("action == NULL");
		return;
    }

```
In all cases the logs appear when the menu is NEXT clicked, consistant with the segfault behavior in GTK 3.23. Function runs twice each time the applet is clicked

GTK 3.22:

Starting panel with "OnDemand" selected, then selecting "2.80GHZ" and clicking again to run this function and create the logs:
```

Jul 12 15:24:51 ubuntu mate-cpufreq-ap[31755]: Governorondemand
Jul 12 15:24:51 ubuntu mate-cpufreq-ap[31755]: Governorondemand
Jul 12 15:24:58 ubuntu mate-cpufreq-ap[31755]: Frequency2626703
Jul 12 15:24:58 ubuntu mate-cpufreq-ap[31755]: Frequency2626703
Jul 12 15:24:58 ubuntu mate-cpufreq-ap[31755]: action == NULL
Jul 12 15:24:58 ubuntu mate-cpufreq-ap[31755]: action == NULL
```

Panel restart with "4.30 GHZ" selected, on first click of the menu and not before:
```

Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: Frequency3834793
Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: Frequency3834793
Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: action == NULL
Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: action == NULL
```

Switch to GTK 3.23.0, July 2nd branch git pull:

Starting panel with "OnDemand" selected, then selecting "2.80GHZ" and clicking again to run this function and create the logs:

```
Jul 12 15:30:48 ubuntu mate-cpufreq-ap[3376]: Governorondemand
Jul 12 15:30:48 ubuntu mate-cpufreq-ap[3376]: Governorondemand
Jul 12 15:31:13 ubuntu mate-cpufreq-ap[3376]: Frequency2651263
Jul 12 15:31:13 ubuntu mate-cpufreq-ap[3376]: Frequency2651263
Jul 12 15:31:13 ubuntu mate-cpufreq-ap[3376]: action == NULL
Jul 12 15:31:13 ubuntu mate-cpufreq-ap[3376]: action == NULL

```

Panel restart with "4.30 GHZ" selected, on first click of the menu and not before:
```

Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: Frequency3834793
Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: Frequency3834793
Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: action == NULL
Jul 12 15:25:55 ubuntu mate-cpufreq-ap[32421]: action == NULL
```
Clicking again, the frequency actuall used in the CPU has changed, giving this:
```

 
Jul 12 15:28:26 ubuntu mate-cpufreq-ap[1895]: Frequency3775833
Jul 12 15:28:26 ubuntu mate-cpufreq-ap[1895]: Frequency3775833
Jul 12 15:28:26 ubuntu mate-cpufreq-ap[1895]: action == NULL
Jul 12 15:28:26 ubuntu mate-cpufreq-ap[1895]: action == NULL

```

lukefromdc:
GTK config.log:

[config.log](https://github.com/mate-desktop/mate-applets/files/2190035/config.log)

GLIB config.log:

[config.log](https://github.com/mate-desktop/mate-applets/files/2190038/config.log)

muktupavels:
Can you repeat without return? Only with 3.22...

```
g_warning ("%s", name);
if (action == NULL) {
     g_warning ("action == NULL");
}
```

lukefromdc:
Rebase for merge

lukefromdc:
That was tested with both GTK versions as logged. I noticed that it was clicking on the menu that brought the previously selected change to the logs by running that function.

lukefromdc:
Confirm: with GTK 3.22 
```

Jul 12 15:57:10 ubuntu mate-cpufreq-ap[24457]: Frequency2353618
Jul 12 15:57:10 ubuntu mate-cpufreq-ap[24457]: action == NULL
Jul 12 15:57:10 ubuntu mate-cpufreq-ap[24457]: Frequency2353618
Jul 12 15:57:10 ubuntu mate-cpufreq-ap[24457]: action == NULL
Jul 12 15:57:19 ubuntu mate-cpufreq-ap[24457]: Frequency2784469
Jul 12 15:57:19 ubuntu mate-cpufreq-ap[24457]: Frequency2784469
Jul 12 15:57:19 ubuntu mate-cpufreq-ap[24457]: action == NULL
Jul 12 15:57:19 ubuntu mate-cpufreq-ap[24457]: action == NULL

```
came up only AFTER clicking the menu a second time

lukefromdc:
I no longer have my GTK 3.22 build logs, but build otions are identical. Here's what I use to configure GTK:

With GTK 3.22 I used

./autogen.sh --enable-wayland-backend --enable-x11-backend --enable-xrandr --enable-xfixes --enable-xkb --disable-maintainer-mode --prefix=/usr  --sysconfdir=/etc libdir=/usr/lib/x86_64-linux-gnu --disable-schemas-compile

and with GTK 3.23 I used 

./autogen.sh --enable-wayland-backend --enable-x11-backend --enable-xrandr --enable-xfixes --enable-xkb --disable-maintainer-mode --prefix=/usr  --sysconfdir=/etc libdir=/usr/lib/x86_64-linux-gnu --disable-schemas-compile --enable-debug=no

because development versions turn on debug warnings by default, and that spams my logs and makes the warnings that users with released builds of GTK will see too hard to find

muktupavels:
```
/*
 * compile with:
 * gcc -Wall `pkg-config --cflags gtk+-3.0` -o action action.c `pkg-config --libs gtk+-3.0`
 */

#include <gtk/gtk.h>

int
main (void)
{
  GtkActionGroup *group;
  GtkAction *action;

  gtk_init (NULL, NULL);

  group = gtk_action_group_new ("my-action-group");
  action = gtk_action_group_get_action (group, "my-action");

  gtk_toggle_action_set_active (GTK_TOGGLE_ACTION (action), TRUE);

  return 0;
}
```

What happens when yo build and run with gtk 3.22 and gtk 3.23?

lukefromdc:
Building with GTK 3.23 installed:
```
luke@ubuntu:~/Desktop$ gcc -Wall `pkg-config --cflags gtk+-3.0` -o action action.c `pkg-config --libs gtk+-3.0`
action.c: In function ‘main’:
action.c:16:3: warning: ‘gtk_action_group_new’ is deprecated [-Wdeprecated-declarations]
   group = gtk_action_group_new ("my-action-group");
   ^~~~~
In file included from /usr/include/gtk-3.0/gtk/gtk.h:250:0,
                 from action.c:6:
/usr/include/gtk-3.0/gtk/deprecated/gtkactiongroup.h:175:17: note: declared here
 GtkActionGroup *gtk_action_group_new                     (const gchar                *name);
                 ^~~~~~~~~~~~~~~~~~~~
action.c:17:3: warning: ‘gtk_action_group_get_action’ is deprecated [-Wdeprecated-declarations]
   action = gtk_action_group_get_action (group, "my-action");
   ^~~~~~
In file included from /usr/include/gtk-3.0/gtk/gtk.h:250:0,
                 from action.c:6:
/usr/include/gtk-3.0/gtk/deprecated/gtkactiongroup.h:195:17: note: declared here
 GtkAction      *gtk_action_group_get_action              (GtkActionGroup             *action_group,
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~
action.c:19:3: warning: ‘gtk_toggle_action_set_active’ is deprecated [-Wdeprecated-declarations]
   gtk_toggle_action_set_active (GTK_TOGGLE_ACTION (action), TRUE);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/gtk-3.0/gtk/deprecated/gtkradioaction.h:36:0,
                 from /usr/include/gtk-3.0/gtk/gtk.h:268,
                 from action.c:6:
/usr/include/gtk-3.0/gtk/deprecated/gtktoggleaction.h:82:18: note: declared here
 void             gtk_toggle_action_set_active        (GtkToggleAction *action,
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
action.c:19:3: warning: ‘gtk_toggle_action_get_type’ is deprecated [-Wdeprecated-declarations]
   gtk_toggle_action_set_active (GTK_TOGGLE_ACTION (action), TRUE);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/gtk-3.0/gtk/deprecated/gtkradioaction.h:36:0,
                 from /usr/include/gtk-3.0/gtk/gtk.h:268,
                 from action.c:6:
/usr/include/gtk-3.0/gtk/deprecated/gtktoggleaction.h:73:18: note: declared here
 GType            gtk_toggle_action_get_type          (void) G_GNUC_CONST;
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~


```

Run with GTK 3.23:

```
(action:3823): dbind-WARNING **: 16:12:26.865: Couldn't register with accessibility bus: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
Segmentation fault (core dumped)

```
Run with GTK 3.22:

```
(action:5426): dbind-WARNING **: 16:14:11.642: Couldn't register with accessibility bus: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
(action:5426): Gtk-CRITICAL **: 16:14:11.642: gtk_toggle_action_set_active: assertion 'GTK_IS_TOGGLE_ACTION (action)' failed

```

muktupavels:
Meh, you just confirmed what I said...

> and with GTK 3.23 I used
>
> ./autogen.sh --enable-wayland-backend --enable-x11-backend --enable-xrandr --enable-xfixes --enable-xkb --disable-maintainer-mode --prefix=/usr --sysconfdir=/etc libdir=/usr/lib/x86_64-linux-gnu --disable-schemas-compile **--enable-debug=no**

https://developer.gnome.org/glib/stable/glib-building.html#extra-configuration-options

> --enable-debug.  Turns on various amounts of debugging support. Setting this to '**no**' **disables** g_assert(), **g_return_if_fail**(), g_return_val_if_fail() and all cast checks between different object types. Setting it to 'minimum' disables only cast checks. Setting it to 'yes' enables runtime debugging. The default is 'minimum' for stable releases, and 'yes' for development snapshots. Note that 'no' is fast, but dangerous as it tends to destabilize even mostly bug-free software by changing the effect of many bugs from simple warnings into fatal crashes. Thus --enable-debug=no should not be used for stable releases of GLib. 

So that is your answer why 3.23 crash and did not crash with 3.22.

lukefromdc:
OK, that tells me to rebuild GTK3.23, I hadn't realized there were multiple levels of debugging support available and simply needed to control what was logged. I will do a new build with "minimum." 

Still, this caused me to find and fix something in the Cpufreq applet that was simply wrong for any GTK version or build, and should crash the applet prior to this commit with any version of GTK3 built with --enable-debug=no, as could happen say with a gentoo user using custom options but who is not a developer.

No application program should depend on GTK being built with a particular option not to crash and especially not to crash with a segfault.  Disabling g_return_if_fail() may also explain all those caja pathbar crashes that depend on the GTK build I have used, as  g_return_if_fail() was just ahead of the offending code. Still needs to be fixed, just won't crash for users with normal GTK builds. I can directly test that with a new GTK build and Caja from master, and if that test confirms things will change the titles of the commits in question. This also permits anyone who can build stable GTK 3.22.30 to test the same issues without needing GTK 3.23 by passing --enable-debug=no in a deliberate test build.

lukefromdc:
OK, this is confirmed: caja from master does not crash on toggling desktop_is_home_dir with a new GTK 3.23.0 build with --enable-debug=minimal but still throws critical warnings from g_object_ref and a failed assert in the local code. Will relabel the related PR.

PR #331, opened 2018-07-15
mateweather-dialog.c: avoid deprecated 'gtk_widget_override_font' v2

sc0w:


sc0w:
> By adding a g_warning for test purposes, I found that the block containing
> gtk_style_context_add_provider_for_screen
> runs seemingly twice (possibly a logging bug as seemingly everything shows up twice in syslog so maybe running only once) and then does not run again.

I have tested writing into a temp file, and that function is called only one time (in other PR, is called every time the dialog is created).

> Albert has concerns about overwriting the default style provider.
> 
> A disadvantage of writing over the default provider is that if we are relying on an undocumented behavior in GTK, that could change with any GTK update, even a point update. At the moment I have it working in GTK 3.23.0 and I assume the original test was with 3.22.30 or similar. These however are not the end of GTK3 updates.

The code of function `gtk_css_provider_get_default` is the same in gtk git master and gtk 3.22:

```
GtkCssProvider *
gtk_css_provider_get_default (void)
{
  static GtkCssProvider *provider;

  if (G_UNLIKELY (!provider))
    {
      provider = gtk_css_provider_new ();
    }

  return provider;
}
```

How it works?

Is very simple, it makes new provider the first time is called, and it gets the same provider next times. (in other PR it makes new provider every time the dialog is created).

lukefromdc:
It looks like this is held to one cssprovider by not freeing    ` provider_added` at the close of the dialog,and I did not find repeat additions of the cssprovider when the dialog was opened multiple times. Am I correct in assuming that on restarting the panel or on removing the applet both this and the cssprovider will be freed?

lukefromdc:
One more option based on this, should solve the freeing problem and any worries about using gtk_css_provider_get_default. I am not committed to any particular version of this, I just want us to end up with the best code, so here's one more option:

https://github.com/mate-desktop/mate-applets/pull/332

sc0w:
> It looks like this is held to one cssprovider by not freeing provider_added at the close of the dialog,and I did not find repeat additions of the cssprovider when the dialog was opened multiple times. Am I correct in assuming that on restarting the panel or on removing the applet both this and the cssprovider will be freed?

Yes, of course, both will be freed.

lukefromdc:
@muktupavels said that GTK itself (the library) cannot "know" when to free its internal instance of  `static GtkCssProvider *provider;` in  `gtk_css_provider_get_default (void)`, while it seems from documentation elsewhere that it will definately be freed on removing the applet or restarting the panel if we set it ourselves in our own application.

I am no expert on static variables and library behavior on closing an application, what does everyone else think?

This PR is consuming an incredible amount of time (I have over 5 hours in it myself over this), and it seems to me this time is best spent if we get a template out of it for handling later instances of dealing with cssproviders in a recurring function. Most of the time, we can add the provider and unref it, but doing that here would require affirmatively removing it, making it anew, re-adding it, and then unrefing the provider. That keeps it to one instance at a time and ensures it is freed, but requires more code to run every time the dialog is opened.

Still, at some point we need to cut the gordian knot and make a choice.

sc0w:
I have tested this same code in pluma, it works as expected, and it doesn't affect the textview of mate-applets. Both are working fine.

lukefromdc:
Will merge this, its a bit simpler than V3 and GNOME would deprecate gtk_css_provider_get_default if they did not intend it to be used.

PR #332, opened 2018-07-16
mateweather-dialog.c: avoid deprecated 'gtk_widget_override_font' v3

lukefromdc:
Use a statically declared cssprovider to hold font data for the forecast tab.

### DIFF
--- a/mateweather/mateweather-dialog.c
+++ b/mateweather/mateweather-dialog.c
@@ -506,6 +506,51 @@ static PangoFontDescription* get_system_monospace_font(void)
     return desc;
 }
 
+static void
+override_widget_font (GtkWidget            *widget,
+                      PangoFontDescription *font)
+{
+    static gboolean provider_added = FALSE;
+    static GtkCssProvider *provider;
+    gchar          *css;
+    gchar          *family;
+    gchar          *weight;
+    const gchar    *style;
+    gchar          *size;
+
+    family = g_strdup_printf ("font-family: %s;", pango_font_description_get_family (font));
+
+    weight = g_strdup_printf ("font-weight: %d;", pango_font_description_get_weight (font));
+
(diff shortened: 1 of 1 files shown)


lukefromdc:
Another way of doing this. This way keeps the code in one function and the call to it, uses a pair of static variables kept the life of the program (should be the applet as it is out-of-process only and can crash w/o taking down the panel) and which are supposed to be automatically freed when the program is terminated and persist until then. One is the  `static gboolean provider_added`  from https://github.com/mate-desktop/mate-applets/pull/331 and the other is the cssprovider itself. This way the cssprovider does not have to be added to the dialog struct, and instead of being freed when the dialog is closed and made again with each new opening of the details dialog stays in memory and is used each time the dialog is opened or updated. 

About freeing/not freeing this:
According to http://eleceng.dit.ie/frank/IntroToC/Memory.html,

"Memory space is alloacted for static variables when a program is loaded from
disk and is only freed when the program terminates."

in mateweather-applet.c there is another instance of a static variable:
static int gw_fault_counter = 0;
It is found in update_finish , it is kept for the lifetime of the program to count a type of error, and it is never explictly free. I am using the same behavior here with a GtkCssProvider and it worked in my testing. If we ever port this to in-process all of these static variables might be something to revisit.

About initialization:

According to
https://stackoverflow.com/questions/13251083/the-initialization-of-static-variables-in-c

"I would add that static variables (or arrays) are classified into two types.
Initialized are the ones that are given value from code at compile time. These are usually stored in DS though this is compiler specific.
The other type is uninitialized statics which are initialized at run time and are stored into BSS segment though again this is compiler specific."

Thus I don't get a compiler error from having a static variable is not initiailized at declaration and gets its value set (once in this case) at runtime by
provider = gtk_css_provider_new()

raveit65:
Runtime test is fine.

PR #333, opened 2018-07-19
weather: Use widget name in cssprovider to limit it to just this applet

lukefromdc:
Will be needed if this is ever ported to in-process. If all of mate-applets including stickynotes was ported to in-process (needed for wayland, GTK4, or any other environment w/o GtkPlug/GtkSocket/Xembed) the font applied here is applied to stickynotes too if the cssprovider is for screen, as in that case the "process" in which the provider is valid becomes the entire panel.

### DIFF
--- a/mateweather/mateweather-dialog.c
+++ b/mateweather/mateweather-dialog.c
@@ -539,7 +539,7 @@ override_widget_font (GtkWidget            *widget,
     gtk_css_provider_load_from_data (provider, css, -1, NULL);
 
     if (!provider_added) {
-        gtk_style_context_add_provider_for_screen (gtk_widget_get_screen (widget),
+        gtk_style_context_add_provider(gtk_widget_get_style_context(GTK_WIDGET(widget)),
                                                    GTK_STYLE_PROVIDER (provider),
                                                    GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
         provider_added = TRUE;
--- a/mateweather/mateweather-dialog.c
+++ b/mateweather/mateweather-dialog.c
@@ -535,11 +535,12 @@ override_widget_font (GtkWidget            *widget,
 
     provider = gtk_css_provider_get_default ();
 
-    css = g_strdup_printf ("textview { %s %s %s %s }", family, weight, style, size);
+    gtk_widget_set_name(GTK_WIDGET(widget), "MateWeatherAppletTextView");
+    css = g_strdup_printf ("#MateWeatherAppletTextView { %s %s %s %s }", family, weight, style, size);
     gtk_css_provider_load_from_data (provider, css, -1, NULL);
 
     if (!provider_added) {
-        gtk_style_context_add_provider(gtk_widget_get_style_context(GTK_WIDGET(widget)),
+        gtk_style_context_add_provider_for_screen (gtk_widget_get_screen (widget),
                                                    GTK_STYLE_PROVIDER (provider),
                                                    GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
         provider_added = TRUE;


muktupavels:
@lukefromdc do you understand what happens in that function?

sc0w:
> the font applied here is applied to stickynotes too

Not here with git master, the font isn't affected in stickynotes if is changed in weather applet, the font is different

Sticky notes uses by the moment the deprecated `gtk_widget_override_font`:

https://github.com/mate-desktop/mate-applets/blob/d1020226236576464be4ab47ec0098a588bef486/stickynotes/stickynotes.c#L556

sc0w:
**Code review:**

you can use simply:

`gtk_style_context_add_provider(gtk_widget_get_style_context(widget),`

instead

`gtk_style_context_add_provider(gtk_widget_get_style_context(GTK_WIDGET(priv->forecast_text)),`

and keep the other parts of code without changes

**Behavior review:**

your PR doesn't work with multiple weather applets, you can test with two weather applets for example, the font is changed only in one applet.

lukefromdc:
@scow, I confirmed your report that only one copy of the applet works if two are in the panel at the same time-and that this is also true if I use Albert's suggestion to use `gtk_style_context_add_provider(gtk_widget_get_style_context(widget`) rather then pass `priv->forecast_text` to `override_widget_font `and use it directly. In both cases, if there are two copies of the weather applet running, only ONE of them gets the font change. So far, only using gtk_style_context_add_provider_for_screen works with two copies of the dialog.  The only reason I resorted to passing `priv->forecast_text` to  `override_widget_font ` was because in my first experiments with the last PR (version 2) I tried switching back to applying the provider for the widget and it did not work at all. Probably made some mistake at that time that did not create a build warning though.

I am going to have to take a look to see what was done about this in gnome-applets, which still has the weather applet. Unless I find something there, I will close this PR and probably be done with this issue entirely. I don't see much likelihood of this being ported to GTK4 or of MATE being ported to Wayland anytime soon, and thus not much likelihood of this applet ever being ported to in-process.

@muktupavels,  I do know what this function does: assembles a cssprovider and adds it either for the screen(last PR as written) or as I just tried to get to work right adds it to the textview widget.  Also, it was you who said back in https://github.com/mate-desktop/mate-applets/pull/325 that it would also affect stickynotes if both were in-process.  Out-of-process no issue.

lukefromdc:
Just removed the use of a third varable to pass and use` priv->forecast_text` to `override_widget_font` but unless this can be applied to all dialogs open and not just the dialog owned by the first instance of the weather applet I will have to close this

lukefromdc:
The gnome-applets gweather applet still uses  `gtk_widget_modify_font ()` which was deprecated all the way back at GTK 3.0, so no help from there. Closing

lukefromdc:
OK, let's try using a widget name instead. I just tested that and it does work, and works across multiple instances of the applet. This will "future-proof" the applet against problems if later ported to in-process (which in the gnome-applets history turns out to be a simple commit).

lukefromdc:
@sc0w, please retest, on my machine this now works with multiple applet instances.

muktupavels:
> I am going to have to take a look to see what was done about this in gnome-applets.

Look again... :D

> Also, it was you who said back in #325 that it would also affect stickynotes if both were in-process. Out-of-process no issue.

I know. Forgot that there was time when gnome-applets was only out-of-process... So right, not a problem to you at least for now. :)

sc0w:
> @scow, I confirmed your report that only one copy of the applet works if two are in the panel at the same time-and that this is also true if I use Albert's suggestion to use gtk_style_context_add_provider(gtk_widget_get_style_context(widget) rather then pass priv->forecast_text to override_widget_fontand use it directly.

That was **my** suggestion :)

> OK, let's try using a widget name instead.

It is a good idea!

It works!

lukefromdc:
@muktupavels ,  does https://gitlab.gnome.org/GNOME/gtk/issues/1226 mean that GNOME will deprecate gtk_css_provider_get_default in GTK 4, and drop it in GTK 5, given that it is not deprecated now, works, and presumably could be used at any time by any application developer using GTK3? I assume nobody in GNOME is going to not only deprecate but remove a not currently deprecated function from GTK3. 

If this is removed, we can use a static variable to handle the cssprovider directly, but this never came up while the discussion was between three versions of the last PR

According to the linked report, the default cssprovider that we are using is empty when first called, and we have found works fine when used here.

muktupavels:
It is already removed from GTK4 and next GTK3 release will warn you!

That default provider was used for fallback styling long time ago and also was added to screen by GTK itself. Things have changed and it does not provide anything. So as it is useless API, it has been removed from GTK4 and deprecation warning added to GTK3.

lukefromdc:
https://github.com/mate-desktop/mate-applets/pull/332 is essentially the exact same code as now offered in 
https://github.com/mate-desktop/mate-applets/pull/334
except that now we are also using the widget name. The only reason I didn't go with that version was that gtk_css_provider_get_default was not deprecated at that time

PR #334, opened 2018-07-20
Avoid newly deprecated gtk_css_provider_get_default

lukefromdc:
This works the exact same way gtk_css_provider_get_default works internally, except that we handle a static cssprovider ourselves. Made necessary by https://gitlab.gnome.org/GNOME/gtk/merge_requests/257

### DIFF
--- a/mateweather/mateweather-dialog.c
+++ b/mateweather/mateweather-dialog.c
@@ -511,7 +511,7 @@ override_widget_font (GtkWidget            *widget,
                       PangoFontDescription *font)
 {
     static gboolean provider_added = FALSE;
-    GtkCssProvider *provider;
+    static GtkCssProvider *provider;
     gchar          *css;
     gchar          *family;
     gchar          *weight;
@@ -532,8 +532,8 @@ override_widget_font (GtkWidget            *widget,
     size = g_strdup_printf ("font-size: %d%s;",
                             pango_font_description_get_size (font) / PANGO_SCALE,
                             pango_font_description_get_size_is_absolute (font) ? "px" : "pt");
-
-    provider = gtk_css_provider_get_default ();
+    if (!provider_added)
+        provider = gtk_css_provider_new ();
 
     gtk_widget_set_name(GTK_WIDGET(widget), "MateWeatherAppletTextView");
     css = g_strdup_printf ("#MateWeatherAppletTextView { %s %s %s %s }", family, weight, style, size);


muktupavels:
Sure, it will work... But why do you need static variables? Why?

lukefromdc:
@muktupavels  I will be very blunt about this: Another team member has clearly decided he wants to restrict this to a single function and sufficiently disliked doing this across multiple functions that he created an alternate PR to the original one. I will not reverse this at the risk of creating problems in the team. If the team breaks up, the code sits statically until it dies by breakage of underlying libraries.

Within this restruction a static variable is necessary to keep it around so we don't either make and unmake the cssprovider multiple times or create many copies of the cssprovider. It is freed when the applet is removed (terminating the application) or the panel restarted. This applet exists only out-of-process in MATE.

It was your decision to deprecate gtk_css_provider_get_default and make this PR necessary, effectively killing version 2 of the original pull request and forcing version 3. I can see why you did this if it is used by nobody else yet has to be maintained by GNOME. None the less, I find gtk_css_provider_get_default in mate-panel, in an old version of Mutter, in mate-desktop, and in pluma in a quick search of my entire development directory. It is not used a lot, but it is used even if not by GNOME and not with the old and broken Raleigh theme.

vkareh:
@muktupavels - how would you restructure the code to avoid the static variable in this case? The only other option I see is calling `gtk_css_provider_get_named` and load it every time for the current theme. But your deprecation of `gtk_css_provider_get_default` suggests using `gtk_css_provider_new`, but I don't think instantiating a brand new GtkCssProvider is a good idea, or is it?

vkareh:
@muktupavels - or are you suggesting that we could just instantiate a new `GtkCssProvider` every time the forecast is opened/updated? I mean, that would work too, but wouldn't that consume more resources unnecessarily? (honest questions, I'm just trying to understand what would make the most sense here)

muktupavels:
@lukefromdc I think in `mutter` it might have been used exactly because it was documented to provide fallback style! If you looked at code then you would see that it is used only if there is no theme name...

@vkareh you create style provider when you need, no? When do you need style provider? Only when dialog is opened, right? So why not create it when you open it and destroy it when dialog is closed?

lukefromdc:
@muktupavels, I had proposed (as you used later in gnome-applets) to put the style provider in the dialog structure and free it in a dispose function (last version of https://github.com/mate-desktop/mate-applets/pull/325) but another team member wanted to keep the replacement for what had been a single line of code calling something GTK deprecated to one function rather than spread it across multiple functions. I had to respect that.

muktupavels:
@lukefromdc You as team member should prefer solution that is best for project...

lukefromdc:
There are many definitions of "best for the project." Team members have to be able to work together, which requires respecting oneanother's opinions.

vkareh:
@muktupavels - fair point. And I guess that the update cycle for this dialog (assuming someone leaves it open indefinitely) is every 30 minutes by default. So releasing memory and re-instantiating at most every half hour should not be an issue.

@lukefromdc - I think at this point things are changing in GTK at such a pace that it might make sense to re-think. The solution you proposed is good and I think we should seriously consider doing that (freeing the provider, as you originally said)

sc0w:
I think is good practice to define the initial value in static variables

I suggest

`static GtkCssProvider *provider = NULL;`

instead

`static GtkCssProvider *provider;`

@muktupavels 
> Sure, it will work... But why do you need static variables? Why?

why `gtk_css_provider_get_default` uses static variable?

This PR works exactly like that function.

sc0w:
remembering the function...

```
GtkCssProvider *
gtk_css_provider_get_default (void)
{
  static GtkCssProvider *provider;

  if (G_UNLIKELY (!provider))
    {
      provider = gtk_css_provider_new ();
    }

  return provider;
}
```

vkareh:
@sc0w - that's a good point, now however, since we own the provider, rather than GTK, we should probably free it up when the dialog is closed. I didn't see any obvious leak warnings with valgrind, but it sounds as though that could be the case. Thoughts?

lukefromdc:
I read somewhere that if you initially define the value of a static variable within a function, it is initialized at compilation not at runtime, and its value must be known at compile time or the build will fail. If it is uninitialized when declared and its value is set later, it is then and only then initialized at runtime

muktupavels:
@vkareh If dialog is open, you can update how many times want. You dont need to re-crate style provider when you update css. You just reuse existing provider!

See how I did that:
https://gitlab.gnome.org/GNOME/gnome-applets/commit/41ea69e46b154e94b4094cd6a6f83ec5eea1570c
(if you see bug / problems, please tell me; more you find - better...)

I have not mentioned it yet, but you dont need to update css every time update button is pressed... You need to update only when setting is changed and/or theme changes that could change font.

@sc0w there is zero reason why it uses static variable these days... And because it is useless it is removed from GTK4 now. But there was time when there was reason, check this commit:
https://gitlab.gnome.org/GNOME/gtk/commit/cdc6e827202059e7aabb112bdc385c0fc826e13e

Do you notice that `gtk_css_provider_get_default` loads default css? And that this provider was added to screen with fallback priority? Then check `gtk_widget_get_default_style`: 
https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-get-default-style

I guess static variable was easiest way to use default provider inside gtk and provide access to applications.

> I think is good practice to define the initial value in static variables

Good practice, but it is not required. For more info:
https://stackoverflow.com/questions/3373108/why-are-static-variables-auto-initialized-to-zero?lq=1

lukefromdc:
A static variable has the runtime of the program as its lifetime. It is created when the program is started and freed when the program terminates. That means for this applet(which exists only out-of-process) it is created when the applet is added to the panel or the panel is started with the applet in it. It is then used for as long as the panel is open with the applet in it. Removing the applet or restarting the panel frees both static variables in this applet.

muktupavels:
@vkareh see - https://github.com/mate-desktop/mate-applets/pull/325#issuecomment-405574846

lukefromdc:
If this was NOT static, it would have to be freed when the dialog was closed and added when the dialog was opened,as in Albert's new work on gnome-applets. It would still have to be written to either when the dialog is updated, or alternately by connecting to a signal for the font changing. If the style provider is also added only when the dialog opens, then the dialog would have to be closed and reopened to apply a font change. That would allow keeping it to one function, though the non-static variables would still have to be freed in a dispose function.

muktupavels:
> If the style provider is also added only when the dialog opens, then the dialog would have to be closed and reopened to apply a font change. 

Why?

lukefromdc:
In order to drop this code and go with a port of the work Albert did in gnome-applets(https://github.com/GNOME/gnome-applets/commit/41ea69e46b154e94b4094cd6a6f83ec5eea1570c)
I would want to have the approval of other team members. Advantage is reducing code delta between the two branches. Disadvantage is this is more complex code at the source level. As I am not the sole maintainer of the project I will not commit to that unless other team members want to go that way.

sc0w:
we can use this simply function with little changes to fix deprecated **gtk_widget_override_font** easily in `stickynotes`

PR #339, opened 2018-08-22
timerapplet: Allow reseting timer when inactive

### LINKED ISSUE #337: Timer applet shouldn't display Finished
Hello,

I think when the Timer applet is finished it should reset to displaying its name instead of displaying 'Finished' until eternity. I have multiple timers related to different activities on my panel and after I used them once I can't see their given name anymore, which makes it hard to tell which one is which. I see no reason to display the text 'Finished' forever after you used it once.

Thank you!
Victoria

vkareh:
Fixes #337

### DIFF
--- a/timerapplet/timerapplet.c
+++ b/timerapplet/timerapplet.c
@@ -72,12 +72,14 @@ static void timer_start_callback (GtkAction *action, TimerApplet *applet);
 static void timer_pause_callback (GtkAction *action, TimerApplet *applet);
 static void timer_stop_callback (GtkAction *action, TimerApplet *applet);
 static void timer_about_callback (GtkAction *action, TimerApplet *applet);
+static void timer_reset_callback (GtkAction *action, TimerApplet *applet);
 static void timer_preferences_callback (GtkAction *action, TimerApplet *applet);
 
 static const GtkActionEntry applet_menu_actions [] = {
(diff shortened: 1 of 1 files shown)


vkareh:
@raveit65 - agreed, the reporter wanted an auto-reset, but I played with that and it was very easy to miss. Another option that worked well enough was that when the timer is Finished, a click on the applet will automatically reset it. I can add that very easily if we want it, but it changes the behavior, so adding a simple action seems harmless for now.

raveit65:
I you like to add an click-on-applet function than i would prefer start/stop the timer.
This is is what i always try to do if i see that applet in panel :)

raveit65:
Btw. is it not possible the display 'Finished timer-name' in panel instead of only 'Finished'?

vkareh:
@raveit65 - true, these are good changes. Here's what I have in mind:

* When timer is inactive, display timer name. On click, start timer.
* When timer is active, display count-down. On click, pause timer.
* When timer is paused, display count-down and pause icon. On click, continue timer.
* When timer is finished, display "Finished %timerName%". On click, reset timer.

@nemkin - do you have opinions on these changes?

nemkin:
This solution sounds good! Thank you! :)

vkareh:
@raveit65 - I've updated this commit to add click interactions with the timer. Please review/test

PR #346, opened 2018-09-27
legal.xml

rbuj:
find . -name legal.xml -exec cp ~/Devel/mate-user-guide/mate-user-guide/C/legal.xml {} \;


lukefromdc:
This is something I really don't know much about: it's the document rather than a build containing it that really needs review, preferably by someone familiar with these legalities

PR #348, opened 2018-09-28
Avoid possible memory leak on battstat applet

rcaridade145:
Reworked update_tooltip on battstat applet to not use `gchar *powerstring`

### DIFF
--- a/accessx-status/applet.c
+++ b/accessx-status/applet.c
@@ -663,7 +663,6 @@ static cairo_surface_t* accessx_status_applet_mousekeys_image(AccessxStatusApple
 	g_object_unref(mouse_pixbuf);
 	g_object_unref(dot_pixbuf);
 
-
 	return surface;
 }
 


muktupavels:
Can you explain this possible leak and how your change fixes it?

rcaridade145:
What i did was rework the code so the variable g_powerstring  was not needed.  It was only needed to apply the constant string to it.

muktupavels:
But where was the memory leak?

rcaridade145:
I believed this could be a memory leak because the gchar was being set to the macro value and then used and not being freed, also not a const var. I refactored the code to not use the var. If  this is wrong I will close the pull request

muktupavels:
`gettext()` does not return newly allocated data. :)

https://www.gnu.org/software/libc/manual/html_node/Translation-with-gettext.html
> Please note that although the return value is char * the returned string must not be changed. This broken type results from the history of the function and does not reflect the way the function should be used. 

So your change does not fix memory leak. Otherwise your change itself is not wrong, simply pull request title is misleading.  It is up to MATE developers to decide what they want to do with this pull request. :)

Simpler change would be to change `gchar *powerstring;` to `const gchar *powerstring;`... To make it clear it should not be freed.

sc0w:
github says:

> This branch cannot be rebased due to conflicts
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

PR #356, opened 2018-10-02
trashapplet manual: Open Trash guimenuitem

rbuj:


lukefromdc:
This works fine, merging as the others have worked fine and not made problems

PR #366, opened 2018-11-21
Make command applet run commands asynchronously

### LINKED ISSUE #163: Mate command applet problem
I use the mate command applet to fetch my ip from ifconfig.me   however it fowls up the applet and locks it up  sometimes bringing the panel down with it.
setps to reproduce
1 set command to curl ifconfig.me
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27904355-mate-command-applet-problem?utm_campaign=plugin&utm_content=tracker%2F517519&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F517519&utm_medium=issues&utm_source=github).
</bountysource-plugin>

yetist:
Fixed #163 , please review, thank you.

### DIFF
--- a/command/command.c
+++ b/command/command.c
@@ -32,6 +32,7 @@
 
 #include <mate-panel-applet.h>
 #include <mate-panel-applet-gsettings.h>
+#include "ma-command.h"
 
 /* Applet constants */
 #define APPLET_ICON    "utilities-terminal"
@@ -54,14 +55,15 @@ typedef struct
     MatePanelApplet   *applet;
 
     GSettings         *settings;
-    GPid               child_pid;
-    gchar             *buffer;
 
     GtkLabel          *label;
     GtkImage          *image;
     GtkBox            *box;
+    MaCommand         *command;
+    GCancellable      *cancellable;
+    gboolean           running;
 
-    gchar             *command;
+    gchar             *cmdline;
     gint               interval;
     gint               width;
 
@@ -71,6 +73,11 @@ typedef struct
 static void command_about_callback (GtkAction *action, CommandApplet *command_applet);
 static void command_settings_callback (GtkAction *action, CommandApplet *command_applet);
 static gboolean command_execute (CommandApplet *command_applet);
+static gboolean command_text_changed (GtkWidget *widget, GdkEvent  *event, gpointer user_data);
+static void interval_value_changed (GtkSpinButton *spin_button, gpointer user_data);
+static void width_value_changed (GtkSpinButton *spin_button, gpointer user_data);
+static void command_async_ready_callback (GObject *source_object, GAsyncResult *res, gpointer user_data);
+static gboolean timeout_callback (CommandApplet *command_applet);
 
 static const GtkActionEntry applet_menu_actions [] = {
     { "Preferences", "document-properties", N_("_Preferences"), NULL, NULL, G_CALLBACK (command_settings_callback) },
@@ -91,22 +98,15 @@ command_applet_destroy (MatePanelApplet *applet_widget, CommandApplet *command_a
         command_applet->timeout_id = 0;
     }
 
+    if (command_applet->cmdline != NULL)
+    {
+        g_free (command_applet->cmdline);
+        command_applet->cmdline = NULL;
+    }
+
     if (command_applet->command != NULL)
     {
-        g_free (command_applet->command);
-        command_applet->command = NULL;
-    }
-
-    if (command_applet->child_pid != 0)
-    {
-        g_spawn_close_pid (command_applet->child_pid);
-        command_applet->child_pid = 0;
-    }
-
-    if (command_applet->buffer != NULL)
-    {
-        g_free (command_applet->buffer);
-        command_applet->buffer = NULL;
+        g_object_unref (command_applet->command);
     }
 
     g_object_unref (command_applet->settings);
@@ -118,18 +118,65 @@ command_about_callback (GtkAction *action, CommandApplet *command_applet)
 {
     const char* authors[] = { "Stefano Karapetsas <stefano@karapetsas.com>", NULL };
 
-    char copyright[] = \
-         "Copyright \xc2\xa9 2015-2018 MATE developers\n"
-         "Copyright \xc2\xa9 2013-2014 Stefano Karapetsas";
-
     gtk_show_about_dialog(NULL,
+                          "title", _("About Command Applet"),
                           "version", VERSION,
-                          "copyright", copyright,
+                          "copyright", _("Copyright \xc2\xa9 2013-2014 Stefano Karapetsas\n"
+                                         "Copyright \xc2\xa9 2015-2019 MATE developers"),
                           "authors", authors,
                           "comments", _("Shows the output of a command"),
                           "translator-credits", _("translator-credits"),
                           "logo-icon-name", APPLET_ICON,
-    NULL );
+                          NULL );
+}
+
+static gboolean
+command_text_changed (GtkWidget *widget, GdkEvent  *event, gpointer user_data)
+{
+    const gchar *text;
+    CommandApplet *command_applet;
+
+    command_applet = (CommandApplet*) user_data;
+    text = gtk_entry_get_text (GTK_ENTRY(widget));
+    if (g_strcmp0(command_applet->cmdline, text) == 0) {
+        return TRUE;
+    }
+
+    if (strlen (text) == 0) {
+        gtk_label_set_text (command_applet->label, ERROR_OUTPUT);
+        return TRUE;
+    }
+
+    g_settings_set_string (command_applet->settings, COMMAND_KEY, text);
+    return TRUE;
+}
+
+static void interval_value_changed (GtkSpinButton *spin_button, gpointer user_data)
+{
+    gint value;
+    CommandApplet *command_applet;
+
+    command_applet = (CommandApplet*) user_data;
+    value = gtk_spin_button_get_value_as_int (spin_button);
+    if (command_applet->interval == value) {
+        return;
+    }
+
+    g_settings_set_int (command_applet->settings, INTERVAL_KEY, value);
+}
+
+static void width_value_changed (GtkSpinButton *spin_button, gpointer user_data)
+{
+    gint value;
+    CommandApplet *command_applet;
+
+    command_applet = (CommandApplet*) user_data;
+    value = gtk_spin_button_get_value_as_int (spin_button);
+    if (command_applet->width == value) {
+        return;
+    }
+
+    g_settings_set_int (command_applet->settings, WIDTH_KEY, value);
 }
 
 /* Show the preferences dialog */
@@ -188,10 +235,13 @@ command_settings_callback (GtkAction *action, CommandApplet *command_applet)
 
     g_signal_connect (dialog, "response", G_CALLBACK (gtk_widget_destroy), dialog);
 
+    g_signal_connect(command, "focus-out-event", G_CALLBACK (command_text_changed), command_applet);
+    g_signal_connect(interval, "value-changed", G_CALLBACK (interval_value_changed), command_applet);
+    g_signal_connect(width, "value-changed", G_CALLBACK (width_value_changed), command_applet);
     /* use g_settings_bind to manage settings */
-    g_settings_bind (command_applet->settings, COMMAND_KEY, command, "text", G_SETTINGS_BIND_DEFAULT);
-    g_settings_bind (command_applet->settings, INTERVAL_KEY, interval, "value", G_SETTINGS_BIND_DEFAULT);
-    g_settings_bind (command_applet->settings, WIDTH_KEY, width, "value", G_SETTINGS_BIND_DEFAULT);
+    g_settings_bind (command_applet->settings, COMMAND_KEY, command, "text", G_SETTINGS_BIND_GET_NO_CHANGES);
+    g_settings_bind (command_applet->settings, INTERVAL_KEY, interval, "value", G_SETTINGS_BIND_GET_NO_CHANGES);
+    g_settings_bind (command_applet->settings, WIDTH_KEY, width, "value", G_SETTINGS_BIND_GET_NO_CHANGES);
     g_settings_bind (command_applet->settings, SHOW_ICON_KEY, showicon, "active", G_SETTINGS_BIND_DEFAULT);
 
     gtk_widget_show_all (GTK_WIDGET (dialog));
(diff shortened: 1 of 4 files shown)


command/command.c
```diff
@@ -96,6 +97,11 @@ command_applet_destroy (MatePanelApplet *applet_widget, CommandApplet *command_a
     }
 
     g_object_unref (command_applet->settings);
+
+    if (command_applet->subprocess != NULL && G_IS_SUBPROCESS(command_applet->subprocess)) {
```
muktupavels: Why `G_IS_SUBPROCESS` is needed here?

command/command.c
```diff
@@ -289,21 +295,71 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+{
+    GSubprocess *subprocess;
+    CommandApplet *command_applet;
+    GInputStream *in;
+    gsize bytes_read;
+    gchar *output;
+    GError *error = NULL;
+
+    command_applet = (CommandApplet *) data;
+    subprocess = G_SUBPROCESS (command_applet->subprocess);
```
muktupavels: Why do you use `G_SUBPROCESS` here?

command/command.c
```diff
@@ -289,21 +295,71 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+        g_warning ("Parse command error: %s\n", error->message);
+        g_clear_error (&error);
+        return FALSE;
     }
-    else
+
+    if (command_applet->subprocess != NULL && G_IS_SUBPROCESS(command_applet->subprocess)) {
+        const gchar *id = g_subprocess_get_identifier (command_applet->subprocess);
+        g_debug ("Process %s is running...\n", id);
+        return TRUE;
```
muktupavels: Should not you also free `argv` here?

command/command.c
```diff
@@ -289,21 +295,71 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+    gsize bytes_read;
+    gchar *output;
+    GError *error = NULL;
+
+    command_applet = (CommandApplet *) data;
+    subprocess = G_SUBPROCESS (command_applet->subprocess);
+
+    if (!g_subprocess_wait_finish (subprocess, result, &error))
+    {
+        g_warning ("Error: %s\n", error->message);
```
muktupavels: Why do you add newline? g_warning, g_debug... does that for you.

command/command.c
yetist: ok, debug code not clean.

command/command.c
yetist: I use source as gsubprocess before, now G_SUBPROCESS not needed.

command/command.c
yetist: Added gstrfreev here.

command/command.c
yetist: They have been deleted.

command/command.c
```diff
@@ -289,21 +295,67 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+    }
+
+    output = g_new0 (gchar, command_applet->width);
+    in = g_subprocess_get_stdout_pipe (subprocess);
+    g_input_stream_read_all (in, output, command_applet->width, &bytes_read, NULL, NULL);
+
+    process_command_output (command_applet, output);
+    g_free (output);
+
+    if (command_applet->subprocess != NULL) {
```
muktupavels: `{` on newline.

command/command.c
```diff
@@ -289,21 +295,67 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+    if (! g_shell_parse_argv (command_applet->command, &argc, &argv, &error))
     {
-        process_command_output (command_applet, output);
+        gtk_label_set_text (command_applet->label, ERROR_OUTPUT);
+        g_clear_error (&error);
+        return FALSE;
     }
-    else
+
+    if (command_applet->subprocess != NULL && G_IS_SUBPROCESS(command_applet->subprocess)) {
```
muktupavels: `G_IS_SUBPROCESS` is not needed here! I think these functions are meant to be used in public API function to check if API user pass correct arguments to functions.

command/command.c
```diff
@@ -289,21 +295,67 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
 command_execute (CommandApplet *command_applet)
 {
     GError *error = NULL;
-    gchar *output = NULL;
-    gint ret = 0;
+    gint argc;
+    gchar **argv;
 
-    if (g_spawn_command_line_sync (command_applet->command, &output, NULL, &ret, &error))
+    if (! g_shell_parse_argv (command_applet->command, &argc, &argv, &error))
```
muktupavels: Remove space after `!`.

command/command.c
```diff
@@ -289,21 +295,67 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+
+    if (!g_subprocess_wait_finish (subprocess, result, &error))
+    {
+        g_clear_error (&error);
+        return;
+    }
+
+    output = g_new0 (gchar, command_applet->width);
+    in = g_subprocess_get_stdout_pipe (subprocess);
+    g_input_stream_read_all (in, output, command_applet->width, &bytes_read, NULL, NULL);
```
muktupavels: This is blocking function, should not you use async version here?

yetist:
Sorry, I tested `dmesg`, but it does not work, the code need changed.

lukefromdc:
Works much better, but one issue remains: some changes of command now display the output of both the new AND that part of the output of the old command not covered by the new output, might be a redraw issue:
setting uname -a and then changing it to curl ifconfig.me will show first an IP address but the text to the right of it from uname -a will still be displayed. Then setting command to dmesg will "clean" this up, and then just setting "uname" or "date" will show a partial rendering of some of the previous text on a second line

yetist:
Maybe now is ok, please test it again, thank you. @lukefromdc

command/command.c
```diff
@@ -289,27 +304,126 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
-    if (g_spawn_command_line_sync (command_applet->command, &output, NULL, &ret, &error))
+    /* command is empty, wait for next timer execution */
+    if (strlen (command_applet->command) == 0)
     {
-        process_command_output (command_applet, output);
+        return TRUE;
     }
-    else
+
+    if (!g_shell_parse_argv (command_applet->command, &argc, &argv, &error))
```
muktupavels: `argc` seems to be unused, it is optional parameter - you can pass `NULL` if you dont need it...

command/command.c
```diff
@@ -289,27 +304,126 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+    return TRUE;
+}
+
+static void
+on_child_exit (GPid child_pid, gint status, gpointer data)
+{
+    CommandApplet *command_applet;
+
+    command_applet = data;
+    g_spawn_close_pid (child_pid);
```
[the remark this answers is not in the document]
yetist: Yes, I use child_pid to know if command is running now at 377 line.

command/command.c
```diff
@@ -289,27 +304,126 @@ process_command_output (CommandApplet *command_applet, gchar *output)
...
+
+    ioc = g_io_channel_unix_new (fd);
+
+    g_io_channel_set_encoding (ioc, NULL, NULL);
+    g_io_channel_set_buffered (ioc, FALSE);
+
+    g_io_channel_set_close_on_unref (ioc, TRUE);
+
+    g_io_add_watch (ioc, cond, func, data);
+    g_io_channel_unref (ioc);
```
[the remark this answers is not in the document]
yetist: OK, now function return void.

lukefromdc:
Last round of changes seems to do it, the applet is now working fine and remains responsive after setting  curl ifconfig.me, and does not have issues changing to something else like dmesg or uname -a after it

raveit65:
With or without PR i doesn't make a different for me.
A little more complex bash commands with a logical operator like this fails.
`nvidia-settings -q GPUCoreTemp | grep -i 0.0 | awk '{print $4}'`

But maybe it is impossible that this can work, so i don't block this PR.

raveit65:
OK.
i can use `bash -c "nvidia-settings -q GPUCoreTemp | grep -i 0.0 | awk '{print $4}'"`
But for some reason i need to reload the panel for getting a new command working.

raveit65:
Weird,
mostly a new command will apply directly, but not always.

muktupavels:
There are still code style problems...

Note that next is what I see in code after your changes (I did not check what happens in master)! With async I see there can be at least theoretical problem... You change setting, timeout is removed, command_execute is called, but it returns because previous command was still running. Guess what now there is no timeout and without user interaction it wont work anymore.

Then one more thing - why do you need call `g_shell_parse_argv` every time? For one command result will be same, no? When you load / update command you could run `g_shell_parse_argv` once and store result to avoid useless work.

I think I would introduce new glib style object, for example, `MaCommand` with following functions.
```
MaCommand *ma_command_new        (const gchar          *command);

void       ma_command_run_async  (MaCommand            *command,
                                  GCancellable         *cancellable,
                                  GAsyncReadyCallback   callback,
                                  gpointer              user_data);

gchar     *ma_command_run_finish (MaCommand            *command,
                                  GAsyncResult         *res,
                                  GError              **error);
```

Then when command changes or is loaded for first time, you cancel previous command if there is one, destroy old command and re-create new.

Timeout should be added always if you have valid command. Perhaps `MaCommand` could implement `GInitable` interface. If command is not valid or parsing fails it could set error and return `NULL`. `ma_command_new` then could be:

```
MaCommand *ma_command_new        (const gchar          *command,
                                  GError              **error);
```

yetist:
> Note that next is what I see in code after your changes (I did not check what happens in master)! With async I see there can be at least theoretical problem... You change setting, timeout is removed, command_execute is called, but it returns because previous command was still running. Guess what now there is no timeout and without user interaction it wont work anymore.

Yes, it does exist.

> Then one more thing - why do you need call `g_shell_parse_argv` every time? For one command result will be same, no? When you load / update command you could run `g_shell_parse_argv` once and store result to avoid useless work.

Every time to call `g_shell_parse_argv` in `command_execute` is required, before the code changes more.
When a user enters a command, every character change causes `command_execute` to be invoked, even during a continuous input process. The current implementation is that if the `text` attribute of `GtkEntry` changes, the `command` changes and `command_execute` is called. Maybe it would be better to modify `comand` and call `command_execute` after `GtkEntry` loses focus. 

> I think I would introduce new glib style object, for example, `MaCommand` with following functions.
> 
> ```
> MaCommand *ma_command_new        (const gchar          *command);
> 
> void       ma_command_run_async  (MaCommand            *command,
>                                   GCancellable         *cancellable,
>                                   GAsyncReadyCallback   callback,
>                                   gpointer              user_data);
> 
> gchar     *ma_command_run_finish (MaCommand            *command,
>                                   GAsyncResult         *res,
>                                   GError              **error);
> ```
> Then when command changes or is loaded for first time, you cancel previous command if there is one, destroy old command and re-create new.
> 
> Timeout should be added always if you have valid command. Perhaps `MaCommand` could implement `GInitable` interface. If command is not valid or parsing fails it could set error and return `NULL`. `ma_command_new` then could be:
> 
> ```
> MaCommand *ma_command_new        (const gchar          *command,
>                                   GError              **error);
> ```
I agree with you, it looks better. If I have enough time, I will try it.

muktupavels:
How often is user changing command? I would say that you should not execute command on every character change, because command will be incomplete, no? Say user types `uptime`? What is point to try to execute `u`, `up`, `upt`, `upti` and `uptim`? Anyway this is outside this merge request...

 `command_execute` execute is timeout callback, it will be 99% called with same command. It should be enough to parse command when command changes.

https://gitlab.gnome.org/GNOME/gnome-applets/merge_requests/6
This is what I got (unfinished). Feel free to re-use and more importantly - find bugs.

yetist:
@raveit65 It's not finished yet.

yetist:
@muktupavels 

Can you complete this [PR](https://gitlab.gnome.org/GNOME/gnome-applets/merge_requests/6)? 
I can learn how to do it in mate-applets by this way, thank you.

muktupavels:
@yetist I don't have time to do it. :(

```
/*
 * compile with:
 * gcc -Wall `pkg-config --cflags gtk+-3.0` -o command ga-command.c main.c `pkg-config --libs gtk+-3.0`
 * valgrind --suppressions=/usr/share/glib-2.0/valgrind/glib.supp --leak-check=full --show-leak-kinds=all -v ./command
 */

#include "ga-command.h"

static guint timeout_id = 0;

static void
run_cb (GObject      *object,
        GAsyncResult *result,
        gpointer      user_data)
{
  GError *error;
  gchar *output;

  error = NULL;
  output = ga_command_run_finish (GA_COMMAND (object), result, &error);

  if (error != NULL)
    {
      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
        g_warning ("%s", error->message);

      g_error_free (error);
      return;
    }

  g_print ("run_cb: `%s`\n", output);
  g_free (output);
}

static gboolean
exit_cb (gpointer user_data)
{
  GMainLoop *loop;

  g_print ("exit_cb\n");

  g_source_remove (timeout_id);
  timeout_id = 0;

  loop = (GMainLoop *) user_data;
  g_main_loop_quit (loop);

  return G_SOURCE_REMOVE;
}

static gboolean
timeout_cb (gpointer user_data)
{
  g_print ("timeout_cb\n");
  return G_SOURCE_CONTINUE;
}

int
main (void)
{
  GMainLoop *loop;
  GError *error;
  GaCommand *command;

  error = NULL;
  command = ga_command_new ("sleep 6s", &error);

  if (error != NULL)
    {
      g_warning ("%s", error->message);
      g_error_free (error);

      return EXIT_FAILURE;
    }

  loop = g_main_loop_new (NULL, FALSE);
  g_timeout_add_seconds (10, exit_cb, loop);

  ga_command_run_async (command, NULL, run_cb, NULL);
  timeout_id = g_timeout_add (500, timeout_cb, NULL);

  g_main_loop_run (loop);

  g_object_unref (command);
  g_main_loop_unref (loop);

  return EXIT_SUCCESS;
}
```

Example with how to use it. The problem is that I think `GaCommand` is not ready yet... While it works I am not sure if I got cancellable part right and I have not verified that there are no memory leaks.

yetist:
This PR is updated, please review it again.

raveit65:
@yetist 
`for pr review` what does that mean?
Can you please squash `for pr review` commit into `Make command applet run commands asynchronously` ?
Or can this be squashed in one commit?
Besides from this it builds and runs fine.
Code-styling looks good.

@muktupavels 
Any concerns left?

muktupavels:
Sorry, I have no time for review...

raveit65:
Ok, i revert the merge and squashed first and second commit together.
https://github.com/mate-desktop/mate-applets/commit/b1006a885a586233e2cc682f4c86d569d7398306
https://github.com/mate-desktop/mate-applets/commit/da78059c5598bdf66828edd1081d3c1ca0f4805b

yetist:
ok, thank you.

PR #369, opened 2018-11-26
Support git.mk to generate .gitignore

yetist:
`git.mk` can autogenerate `.gitignore` file for autotool-based project, but `Makefile.am` should include it.

How to use:
```bash
$ git config alias.down-git-mk $'!wget https://github.com/behdad/git.mk/raw/master/git.mk'
$ git down-git-mk    # will download git.mk file.
$ make -f git.mk 
$ # please run ./autogen.sh or ./configure to generate Makefile.
$ make gitignore # will generate .gitignore file
$ git status # will see clean output now.
```

### DIFF
--- a/git.mk
+++ b/git.mk
@@ -1,0 +1,400 @@
+# git.mk, a small Makefile to autogenerate .gitignore files
+# for autotools-based projects.
+#
+# Copyright 2009, Red Hat, Inc.
+# Copyright 2010,2011,2012,2013 Behdad Esfahbod
+# Written by Behdad Esfahbod
+#
+# Copying and distribution of this file, with or without modification,
+# is permitted in any medium without royalty provided the copyright
+# notice and this notice are preserved.
+#
+# The latest version of this file can be downloaded from:
+GIT_MK_URL = https://raw.githubusercontent.com/behdad/git.mk/master/git.mk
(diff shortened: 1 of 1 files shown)


muktupavels:
Why git.mk is not added? Do you expect that everyone will manually download it?

lukefromdc:
I do not know what this does or is used for so someone else will need to test it

yetist:
> Why git.mk is not added? Do you expect that everyone will manually download it?

I want to add git.mk, but it looks like the mate desktop team doesn't like it. Because I found that in mate-session-manager, there should have been git.mk, but it does not exist.
https://github.com/mate-desktop/mate-session-manager/blob/master/Makefile.am#L41

raveit65:
I think it was a personal decision of previous MATE founders not to use such a file in git repos.
Personal i am fine with adding those file.
Thoughts?
@lukefromdc @monsta @vkareh @sc0w

lukefromdc:
I have no idea, as I never build from the same directory that I push to git, always making a copy(which is trivial to do)

raveit65:
@yetist 
Can you add the `git.mk` to PR please.
I don't think it is a big deal , as nobody else did answer to my question.

PR #383, opened 2019-03-14
Use g_list_free_full() convenience function

rbuj:
```C
void
g_list_free_full (GList *list,
                  GDestroyNotify free_func);
```
Convenience method, which frees all the memory used
by a GList, and calls `free_func` on every element's
data.

https://developer.gnome.org/glib/stable/glib-Doubly-Linked-Lists.html#g-list-free-full


muktupavels:
Same here, one commit per applet might be better.

PR #406, opened 2019-06-14
treat output as utf8 when limiting width. fixes #374

### LINKED ISSUE #374: Command applet truncates in the middle of a unicode code point.
It looks like [the code](https://github.com/mate-desktop/mate-applets/blob/7a8e4de6b0292d33a1bfe50a644c00e60ebb7b97/command/command.c#L283) is treating the output as a byte string instead of a unicode string.

#### Expected behaviour
Applet shows `…` (preferably) or an empty string (less preferably).

#### Actual behaviour
Applet shows a replacement character.

1. Create a command applet.
2. Set the command to `printf '%s' '…'`.
3. Set the maximum width to `1`.

TheJonny:
should it be utf8 always?

### DIFF
--- a/command/command.c
+++ b/command/command.c
@@ -23,9 +23,9 @@
  */
 
 #include <config.h>
-#include <string.h>
 
 #include <glib.h>
+#include <gmodule.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
 #include <gtk/gtk.h>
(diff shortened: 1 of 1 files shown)


TheJonny:
please wait, as i don't know exactly, what happens when invalid utf8 is passed

raveit65:
Can you rename first commit. We don' add links in commit header. Please use commit body for that.
Better....
```
 treat output as utf8 when limiting width.

 fixes #374 
```

raveit65:
@vkareh 
Thanks for review, but why not merging it as second reviewer?
This is not a complicated change and easy to confirm.

PR #409, opened 2019-07-10
multiload: Use /proc/diskstats for NVMe drives

### LINKED ISSUE #290: Nvme disk activity missing from resource monitor applet
Expected behaviour

System Monitor shows all disk activity, including disk nvme devices.

Actual behaviour

No disk activity is shown for nvme disk devices.

Steps to reproduce the behaviour

Add System Monitor panel to Mate desktop panel.
Right click, preferences, enable hard disk monitoring
Use the disk and notice that disk activity will always be missing
MATE general version

Tested and missing versions 1.14.x through 1.18.x

Package version

1.18.4

Linux Distribution

Debian

Link to downstream report of your Distribution

Don't see anything in Debian downstream, since this appears to be an upstream issue. Ubuntu has some users talking about it?

https://ubuntu-mate.community/t/system-monitor-panel-applet/7841

vkareh:
Since glibtop does not support NVMe drives, we rely on /proc/diskstats
to tell use load information for NVMe drives. If diskstats is not
present it returns to glibtop as a fallback.

This also adds a preference to toggle this behavior (default is glibtop).

Fixes #290

### DIFF
--- a/multiload/linux-proc.c
+++ b/multiload/linux-proc.c
@@ -100,14 +100,14 @@ GetDiskLoad (int Maximum, int data [3], LoadGraph *g)
     static guint64 lastread = 0, lastwrite = 0;
     static AutoScaler scaler;
 
-    glibtop_mountlist mountlist;
-    glibtop_mountentry *mountentries;
     guint i;
     int max;
+    gboolean nvme_diskstats;
 
     guint64 read, write;
     guint64 readdiff, writediff;
 
+    nvme_diskstats = g_settings_get_boolean (g->multiload->settings, "diskload-nvme-diskstats");
 
     if(first_call)
     {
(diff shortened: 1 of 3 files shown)


raveit65:
https://github.com/orgs/mate-desktop/teams/core-team/discussions/17

raveit65:
I know that old code checks only for /dev/ /media mountpoints.
But what is with /mnt ?

raveit65:
When i copy data (15.7 GB) from single nvme drive `/dev/nvme4n1p3` to a linux software raid=0 of 4 nvme-drives `/dev/md/fed30`, than the tooltip of the graph from the applet shows me 100% disk usage for 10 secs.
But the graph shows me only a short peak when the copy action was started.

Normally i don't stare on the disk graph, so i am not sure if this is normal behaviour with magnetic hard disk too?

PR #415, opened 2019-08-25
 timerapplet.c: remove reductant condition

shlyakpavel:


cwendling:
The build failure seems unrelated to this change, but is concerning (maybe an incompatible library version?)

shlyakpavel:
@cwendling build failure is explained here https://github.com/mate-desktop/mate-applets/issues/416

PR #428, opened 2019-09-23
build: fix building cpufreq-applet for s390x arch

raveit65:
Don't stop building when cpufreq.h isn't available.


sc0w:
I confirm, the build is ok if `cpufreq.h` is missing with this change, I can make `.deb` packages sucessfully

PR #433, opened 2019-12-13
make the dbus directory configurable

eworm-de:


monsta:
I guess we need to change `DBUS_REQUIRED` to require 1.10 too?

configure.ac
```diff
@@ -134,6 +134,17 @@ fi
...
 AC_SUBST(LIBNOTIFY_LIBS)
 
+dnl -- Check for D-Bus -----------------------------------------
+AC_ARG_WITH(dbus-sys,
+        [AC_HELP_STRING([--with-dbus-sys=<dir>],
+        [where D-BUS system.d directory is])])
+if ! test -z "$with_dbus_sys" ; then
+        DBUS_SYS_DIR="$with_dbus_sys"
+else
+        DBUS_SYS_DIR='${sysconfdir}/dbus-1/system.d'
```
monsta: Should `datadir` be here? Like in https://github.com/mate-desktop/mate-settings-daemon/pull/304.

eworm-de: The new dbus default is to place it in `datadir`, however using `sysconfdir` is the safe default to make sure it works with older dbus. I am fine with either of both.

If you want to use `datadir` you should consider to change in a separate commit as it changes the current behavior.

monsta: I think 1.10 is old enough to require it as a minimum version.

eworm-de: Fine then. 👍

PR #437, opened 2020-01-29
Fix build using gcc 10 -fno-common flag

rbuj:
```
/usr/bin/ld: drive-list.o:/home/robert/mate-applets/drivemount/drive-list.h:68: multiple definition of `settings'; drivemount.o:/home/robert/mate-applets/drivemount/drive-list.h:68: first defined here
```

### DIFF
--- a/drivemount/drive-list.c
+++ b/drivemount/drive-list.c
@@ -32,6 +32,8 @@
 
 G_DEFINE_TYPE (DriveList, drive_list, GTK_TYPE_GRID);
 
+GSettings *settings;
+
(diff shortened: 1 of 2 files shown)


monsta:
@rbuj: interesting, that `GSettings` variable is actually used only in `drivemount.c`, but is declared and freed in `drive-list.c`. Maybe this could be simplified?

rbuj:
@monsta Do you mean something similar to https://github.com/mate-desktop/mate-applets/pull/440?

PR #438, opened 2020-01-29
I18n tooltip text: nothing to mount

rbuj:
Before:

before

After:

after

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -245,7 +245,7 @@ AC_DEFINE_UNQUOTED(GETTEXT_PACKAGE, "$GETTEXT_PACKAGE", [Gettext Package])
 AC_DEFINE_DIR(MATELOCALEDIR, "${datadir}/locale", [locale directory])
 
 # this is the directory where the *.{mo,gmo} files are installed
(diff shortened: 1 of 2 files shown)


monsta:
@rbuj: I could have missed it, but is `matelocaledir` variable actually used in any `Makefile.am`? I only found `MATELOCALEDIR` being used in the netspeed applet.

rbuj:
`DATADIRNAME` has been removed from gettext 0.20. 

https://github.com/coreutils/gnulib/commit/46133252579ffd78438ebaa2761ec19ecd1e1297#diff-7b950a0fa3c34d0aa7044293296b3126L364-L366

monsta:
No, I meant the actual usage of `matelocaledir` variable...

rbuj:
If the somethinglocaldir variable is set, such as setting matelocaledir, it will be used for setting where to install the translation files.

monsta:
I thought it was set via `MATELOCALEDIR`? Also `-DMATELOCALEDIR` should be used in `Makefile.am`, and `bindtextdomain` should be called in the code?

monsta:
Ok, let's see what happens if we drop that... in #445.

PR #445, opened 2020-03-04
build: drop unused locale stuff

monsta:
Follow-up to #438.
How to test: check that all applets are still translated as before.


raveit65:
`drivemount: i18n of 'nothing to mount' tooltip text`  is in 1.24 release. 
This needs to be cherry-picked to 1.24 branch, or not?

rbuj:
@monsta done. Thank you very much for the reminder! I was overlooked ;-)

PR #447, opened 2020-03-25
Generate missing 256x256 and 64x64 icons from svg files

rbuj:


raveit65:
Same result with mate-eyes cupufreq icons. Only installing the new SVGs fixes the problem.
But we can merge the PR like it is to avoid unneeded work to remove the few new PNGs from PR, if you want.

PR #452, opened 2020-04-08
multiload: declaration of ‘screen’ shadows a previous local

rbuj:


sc0w:
in the same function I see:
```
		gtk_window_set_screen (GTK_WINDOW (dialog),
				       gtk_widget_get_screen (GTK_WIDGET (ma->applet)));
```

you can use the variable `screen` there too

rbuj:
@sc0w done.

PR #454, opened 2020-04-16
build: use dbus-glib only when required

### LINKED ISSUE #453: dbus-glib deprecation
#### Expected behaviour

cpufreq applets builds without the need of dbus-glib

#### Actual behaviour

dbus-glib is a build requirement if HAVE_POLKIT is set.

...

#### MATE general version

1.24

#### Package version

1.24.0

#### Linux Distribution

Debian

#### Link to downstream report of your Distribution

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=955919

rbuj:
Closes #453

### DIFF
--- a/cpufreq/src/Makefile.am
+++ b/cpufreq/src/Makefile.am
@@ -6,8 +6,7 @@ SUBDIRS = $(selector_SUBDIR)
 
 AM_CPPFLAGS = \
 	-DCPUFREQ_MENU_UI_DIR=\""$(datadir)/mate/ui"\" 	\
-	$(MATE_APPLETS4_CFLAGS)				\
-	$(MATEDESKTOP_CFLAGS)
+	$(MATE_APPLETS4_CFLAGS)
 
 if HAVE_DBUS
 AM_CPPFLAGS += $(DBUS_CFLAGS)
@@ -37,7 +36,6 @@ mate_cpufreq_applet_CFLAGS =		\
 
 mate_cpufreq_applet_LDADD =		\
 	$(MATE_APPLETS4_LIBS)   	\
-	$(MATEDESKTOP_LIBS)		\
 	$(LIBCPUFREQ_LIBS)
 
 if HAVE_DBUS
--- a/cpufreq/src/Makefile.am
+++ b/cpufreq/src/Makefile.am
@@ -8,10 +8,6 @@ AM_CPPFLAGS = \
 	-DCPUFREQ_MENU_UI_DIR=\""$(datadir)/mate/ui"\" 	\
 	$(MATE_APPLETS4_CFLAGS)
 
-if HAVE_DBUS
-AM_CPPFLAGS += $(DBUS_CFLAGS)
-endif
-
 libexec_PROGRAMS = mate-cpufreq-applet
 
 if HAVE_LIBCPUFREQ
@@ -37,9 +33,5 @@ mate_cpufreq_applet_CFLAGS =		\
 mate_cpufreq_applet_LDADD =		\
 	$(MATE_APPLETS4_LIBS)   	\
 	$(LIBCPUFREQ_LIBS)
-
-if HAVE_DBUS
-mate_cpufreq_applet_LDADD += $(DBUS_LIBS)
-endif
 
 -include $(top_srcdir)/git.mk
--- a/cpufreq/src/Makefile.am
+++ b/cpufreq/src/Makefile.am
@@ -6,9 +6,12 @@ SUBDIRS = $(selector_SUBDIR)
 
 AM_CPPFLAGS = \
 	-DCPUFREQ_MENU_UI_DIR=\""$(datadir)/mate/ui"\" 	\
-	$(MATE_APPLETS4_CFLAGS) 				\
-	$(MATEDESKTOP_CFLAGS) 				\
-	$(DBUS_CFLAGS)
+	$(MATE_APPLETS4_CFLAGS)				\
+	$(MATEDESKTOP_CFLAGS)
+
+if HAVE_DBUS
+AM_CPPFLAGS += $(DBUS_CFLAGS)
+endif
 
 libexec_PROGRAMS = mate-cpufreq-applet
 
@@ -35,7 +38,10 @@ mate_cpufreq_applet_CFLAGS =		\
 mate_cpufreq_applet_LDADD =		\
 	$(MATE_APPLETS4_LIBS)   	\
 	$(MATEDESKTOP_LIBS)		\
-	$(LIBCPUFREQ_LIBS)		\
-	$(DBUS_LIBS)
+	$(LIBCPUFREQ_LIBS)
+
+if HAVE_DBUS
+mate_cpufreq_applet_LDADD += $(DBUS_LIBS)
+endif
 
 -include $(top_srcdir)/git.mk
--- a/cpufreq/src/cpufreq-selector.c
+++ b/cpufreq/src/cpufreq-selector.c
@@ -21,7 +21,7 @@
 #include <sys/sysinfo.h>
 
 #ifdef HAVE_POLKIT
-#include <dbus/dbus-glib.h>
+#include <gio/gio.h>
 #endif /* HAVE_POLKIT */
 
 #include "cpufreq-selector.h"
@@ -30,7 +30,8 @@ struct _CPUFreqSelector {
 	GObject parent;
 
 #ifdef HAVE_POLKIT
-	DBusGConnection *system_bus;
+	GDBusConnection *system_bus;
+	GDBusProxy      *proxy;
 #endif /* HAVE_POLKIT */
 };
 
@@ -46,7 +47,8 @@ cpufreq_selector_finalize (GObject *object)
 	CPUFreqSelector *selector = CPUFREQ_SELECTOR (object);
 
 #ifdef HAVE_POLKIT
-	selector->system_bus = NULL;
+	g_clear_object (&selector->proxy);
+	g_clear_object (&selector->system_bus);
 #endif /* HAVE_POLKIT */
 
 	G_OBJECT_CLASS (cpufreq_selector_parent_class)->finalize (object);
@@ -84,7 +86,7 @@ typedef enum {
 
 typedef struct {
 	CPUFreqSelector *selector;
-	
+
 	CPUFreqSelectorCall call;
 
 	guint  cpu;
@@ -93,9 +95,6 @@ typedef struct {
 
 	guint32 parent_xid;
 } SelectorAsyncData;
-
-static void selector_set_frequency_async (SelectorAsyncData *data);
-static void selector_set_governor_async  (SelectorAsyncData *data);
 
 static void
 selector_async_data_free (SelectorAsyncData *data)
@@ -114,36 +113,52 @@ cpufreq_selector_connect_to_system_bus (CPUFreqSelector *selector,
 	if (selector->system_bus)
 		return TRUE;
 
-	selector->system_bus = dbus_g_bus_get (DBUS_BUS_SYSTEM, error);
+	selector->system_bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, NULL, error);
 
 	return (selector->system_bus != NULL);
 }
 
-static void
-dbus_set_call_notify_cb (DBusGProxy     *proxy,
-			 DBusGProxyCall *call,
-			 gpointer        user_data)
+
+static gboolean
+cpufreq_selector_create_proxy (CPUFreqSelector  *selector,
+                               GError          **error)
 {
-	SelectorAsyncData *data;
-	GError            *error = NULL;
+	if (selector->proxy)
+		return TRUE;
 
-	data = (SelectorAsyncData *)user_data;
-	
-	if (dbus_g_proxy_end_call (proxy, call, &error, G_TYPE_INVALID)) {
-		selector_async_data_free (data);
-		return;
+	selector->proxy = g_dbus_proxy_new_sync (selector->system_bus,
+	                                         G_DBUS_PROXY_FLAGS_NONE,
+	                                         NULL,
+	                                         "org.mate.CPUFreqSelector",
+	                                         "/org/mate/cpufreq_selector/selector",
+	                                         "org.mate.CPUFreqSelector",
+	                                         NULL,
+	                                         error);
+
+	return (selector->proxy != NULL);
+}
+
+static void
+selector_setter_cb (GObject      *source,
+                    GAsyncResult *result,
+                    gpointer      user_data)
+{
+	GDBusProxy *proxy = G_DBUS_PROXY (source);
+	SelectorAsyncData *data = (SelectorAsyncData *)user_data;
+	GError *error = NULL;
+
+	g_dbus_proxy_call_finish (proxy, result, &error);
+	if (error != NULL) {
+		g_warning ("%s", error->message);
+		g_clear_error (&error);
 	}
-
 	selector_async_data_free (data);
-	g_warning ("%s", error->message);
-	g_error_free (error);
 }
 
 static void
 selector_set_frequency_async (SelectorAsyncData *data)
 {
-	DBusGProxy *proxy;
-	GError     *error = NULL;
+	GError *error = NULL;
 
 	if (!cpufreq_selector_connect_to_system_bus (data->selector, &error)) {
 		g_warning ("%s", error->message);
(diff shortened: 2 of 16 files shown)


sc0w:
if this closes https://github.com/mate-desktop/mate-applets/issues/453 , `dbus-glib` packages can be removed at travis

rbuj:
> if this closes #453 , `dbus-glib` packages can be removed at travis

You can disable PolicyKit support using `--disable-polkit` option.

```shell
$ ./autogen.sh --prefix=/usr --disable-polkit
<cut>
mate-applets-1.24.0 configure summary:

    Prefix:                        /usr
    Source code location:          .
    Compiler:                      gcc
    Compiler flags:                -g -O2
    Compiler warnings:             -Wall -Wmissing-prototypes

    Building:
        - accessx-status           true
        - battstat                 yes
        - charpick                 always
            - gucharmap support    yes
        - cpufreq                  yes
            - building selector    yes
            - using PolicyKit      no
            - enabling suid bit    yes
        - drivemount               always
        - geyes                    always
        - mateweather              true
        - multiload                true
        - netspeed                 true
            - iwlib support        yes
        - stickynotes              yes
        - timerapplet              yes
        - trashapplet              always

    Using UPOWER:                  yes
    Using libnotify:               yes
    Enabling IPv6:                 yes

Now type `make' to compile mate-applets
```

raveit65:
But polkit is needed for password query before changing frequency.

raveit65:
For what was dbus-glib used?

rbuj:
dbus-glib is not used in mateweather:
```
$ git grep dbus-glib -- ':*.c' ':*.h'
$ git grep dbus_g -- ':*.c' ':*.h'
```

raveit65:
```
[rave@mother mate-applets]$ git grep dbus-glib -- ':*.c' ':*.h'
cpufreq/src/cpufreq-selector.c:#include <dbus/dbus-glib.h>
cpufreq/src/cpufreq-selector/cpufreq-selector-service.c:#include <dbus/dbus-glib-lowlevel.h>
cpufreq/src/cpufreq-selector/cpufreq-selector-service.h:#include <dbus/dbus-glib.h>
cpufreq/src/cpufreq-utils.c:#include <dbus/dbus-glib.h>
```
But we need it for cpu-freq applet. 
Here it is used to  build cpufreq-selector with polkit?

raveit65:
So, you want to allow that a unprivileged user can change cpu FREQUENCY and GOVERNOR of the system with the `suid bit` without typing a password?
I think this shouldn't be allowed on a linux multi-user-system.
Is other words we lost cpufreq-applet without `dbus-glib` or replacing it with GDbus.

raveit65:
see 
https://gitlab.gnome.org/GNOME/gnome-applets/-/commit/ed5418eed57cb00976603923d2ccf2b1b8acddf4
and
https://gitlab.gnome.org/GNOME/gnome-applets/-/commit/b24ea7511461bc9ab1f97ff77c45963364cb44db

rbuj:
@raveit65 You cannot change your CPU frequency when you build it without polkit support, i.n. the drop-down menu for selecting cpu frequency is not displayed on clicking.
```
./autogen.sh --prefix=/usr --disable-polkit && make && sudo make install
```
The screenshot below shows the drop-down menu for changing the CPU frequency when polkit has been enabled.
Screenshot at 2020-05-17 12-03-24

raveit65:
@rbuj 
Yes, but i want to build it with `--enable-polkit`. It seems that this isn't possible with this PR.
As i said, allowing that a unprivileged user can change cpu FREQUENCY and GOVERNOR of the system with the suid bit without typing a password, is bad, and i don't want that for fedora.

And the password query pops up after you select another frequency or governor.

raveit65:
Please test for yourself for better understanding what i like to say.
On a system ***without*** `dbus-glib` but ***with*** `polkit-devel` installed, config summary is always
```
            - cpufreq                  yes
            - building selector    yes
            - using PolicyKit      no
            - enabling suid bit    yes
```
So, you can`t remove
```
PKG_CHECK_MODULES(DBUS, dbus-1 >= $DBUS_REQUIRED dbus-glib-1 >= $DBUS_GLIB_REQUIRED, HAVE_DBUS=yes, HAVE_DBUS=no)
```
configure.ac.
`dbus-glib` needs to be replaced with `GDbus` like in gnome-panel.

rbuj:
> Please test for yourself for better understanding what i like to say.
> On a system _**without**_ `dbus-glib` but _**with**_ `polkit-devel` installed, config summary is always
> 
> ```
>         - cpufreq                  yes
>             - building selector    yes
>             - using PolicyKit      no
>             - enabling suid bit    yes
> ```
> 
> So, you can`t remove
> 
> ```
> PKG_CHECK_MODULES(DBUS, dbus-1 >= $DBUS_REQUIRED dbus-glib-1 >= $DBUS_GLIB_REQUIRED, HAVE_DBUS=yes, HAVE_DBUS=no)
> ```

you can remove the previous line since `dbus-glib` is already on
https://github.com/mate-desktop/mate-applets/blob/7508532f97226b043410fb611d62d4f34b1052c5/configure.ac#L120

> checking for POLKIT... no
> configure: error: PolicyKit support explicitly requested but dependencies not found

POLKIT requires `dbus-glib`, so what is wrong?

sc0w:
I think this PR doesn't fix https://github.com/mate-desktop/mate-applets/issues/453 because the reporter says:

> Actual behaviour
> dbus-glib is a build requirement if HAVE_POLKIT is set.

and this behavior is the same with this PR

muktupavels:
> POLKIT requires dbus-glib, so what is wrong?

Do you know what `PKG_CHECK_MODULES` does?

I would guess that `dbus-glib-1` is leftover after porting to PolicyKit 1.0. That line does not say that polkit requires anything. It just checks if listed modules are available and sets `*_CFLAGS`  and `*_LIBS`.

rbuj:
> > POLKIT requires dbus-glib, so what is wrong?
> 
> Do you know what `PKG_CHECK_MODULES` does?
> 
> I would guess that `dbus-glib-1` is leftover after porting to PolicyKit 1.0. That line does not say that polkit requires anything. It just checks if listed modules are available and sets `*_CFLAGS` and `*_LIBS`.

look at error prompt

muktupavels:
Please learn how things works...

rbuj:
@muktupavels I will ignore your last comment...

raveit65:
But cpufreq-selector.c needs `dbus-glib` when compiling with `--enable-polkit`.
See https://github.com/mate-desktop/mate-applets/blob/master/cpufreq/src/cpufreq-selector.c#L23
```
#ifdef HAVE_POLKIT
#include <dbus/dbus-glib.h>
#endif /* HAVE_POLKIT */
```
From `cpufreq-selector.c#L79` to `cpufreq-selector.c#L243`
we need dbus-glib.
Why you want to run `configure --enable-polkit` without checking for dbus-glib?

> > checking for POLKIT... no
>> configure: error: PolicyKit support explicitly requested but dependencies not found

> POLKIT requires `dbus-glib`, so what is wrong?

The message says only that a dependency is missing when `polkit-devel` is installed, but not that dbus-glib is missing and users have no info what is missing.

raveit65:
> you can remove the previous line since dbus-glib is already on

No, i can remove dbus-glib-devel from my installation and polkit-devel is installed.
So  dbus-glib isn't already on here :)

muktupavels:
You can ignore, but that does not change fact that apparently you do not know how things works (based on your comments)...

You pointing me to error message does not make polkit to depend on dbus-glib. If you would understand how things works you would not write such comment...

That error line just says that polkit support was enabled - automatically or by using `--enable-polkit`. It checked if your system has `polkit-gobject-1` and `dbus-glib-1` and if it failed to find one of them you get your error. It does not make polkit to depend on dbus-glib...

I could change that line `PKG_CHECK_MODULES(POLKIT, polkit-gobject-1 dbus-glib-1 my-random-lib, HAVE_POLKIT=yes, HAVE_POLKIT=no)` and it will give you that error message always because `my-random-lib` module/library does not exist.

I told you that it is probably leftover from porting to "PolicyKit 1.0". If you would have tried to find and look at related commit you would have seen that in past polkit API did use `DBusConnection`. That is why it `dbus-glib-1` is there. You would not be able to use polkit API without dbus-glib...

So speaking about polkit you could have small cleanup commit that removes `dbus-glib-1` from `POLKIT` modules list because polkit itself no longer needs `dbus-glib`.

Now look at lines you removed:
```
dnl -- check for dbus (required for cpufreq-selector)
PKG_CHECK_MODULES(DBUS, dbus-1 >= $DBUS_REQUIRED dbus-glib-1 >= $DBUS_GLIB_REQUIRED, HAVE_DBUS=yes, HAVE_DBUS=no)
```

It says for what it is needed.

raveit65:
@muktupavels 
Interesting what you said about the leftover in policykit-1.0.
But why you ported cpufreq from gnome-applets from dbus-glib to GDbus?
https://gitlab.gnome.org/GNOME/gnome-applets/-/commit/ed5418eed57cb00976603923d2ccf2b1b8acddf4
and
https://gitlab.gnome.org/GNOME/gnome-applets/-/commit/b24ea7511461bc9ab1f97ff77c45963364cb44db


Mainly with the first commit you dropped dbus-gllib dependencies for cpufreq-selector.c and cpufreq-utils.c
And second commit drops the rest of dbus-gllib dependencies.

rbuj:
This PR doesn't [migrate to GDBus](https://www.freedesktop.org/software/gstreamer-sdk/data/docs/latest/gio/ch30.html).

raveit65:
> This PR doesn't migrate to GDBus.

Exactly, this is missing to solve https://github.com/mate-desktop/mate-applets/issues/453

raveit65:
Your changes for ` mateweather/Makefile.am` are OK.

But i don't understand you motivation to remove dbus-glib PKG_CHECK_MODULES, DBUS_CFLAGS and DBUS_LIBS for cpufreg when it is needed to build with `polkit`?
`dbus-glib` is used to build code of cpufreq-selector, cpufreq-selector-service.c and cpufreq-utils.c
It is needed to use the cpufreq-selector with a password query.

Again the `suid bit` option isn't really a solution for a multiuser OS, IHMO.

...ok, i start to repeating myself....i will stop now.

muktupavels:
@raveit65 leftover is in that line that checks for POLKIT. Polkit can be used without `dbus-glib`...

What I mean is that you can have `PKG_CHECK_MODULES(POLKIT, polkit-gobject-1 >= $POLKIT_REQUIRED, HAVE_POLKIT=yes, HAVE_POLKIT=no)` and it will work same.

But looking at code I see one problem. @rbuj forgot to mention "small" detail - all `dbus-glib` code is behind `HAVE_POLKIT` ifdefs. So I guess change itself is right. If polkit is disabled no `dbus-glib` related code will be compiled.

Of course this still does not fix linked issue and polkit itself does not require/depend on `dbus-glib-1`.

And it just weird to see `HAVE_POLKIT` in `cpufreq-utils.c` and `cpufreq-selector.c` when it actually does not use polkit there... Basically bad choice for define, `USING_POLKIT` would be more appropriate.

And building applet without polkit is useless:
https://github.com/mate-desktop/mate-applets/blob/master/cpufreq/src/cpufreq-utils.c#L165

In that case it search for `cpufreq-selector`, but you have `mate-cpufreq-selector`... So clearly no one is using applet without polkit.

cpufreq/src/Makefile.am
```diff
@@ -35,7 +38,10 @@ mate_cpufreq_applet_CFLAGS =		\
 mate_cpufreq_applet_LDADD =		\
 	$(MATE_APPLETS4_LIBS)   	\
 	$(MATEDESKTOP_LIBS)		\
-	$(LIBCPUFREQ_LIBS)		\
-	$(DBUS_LIBS)
+	$(LIBCPUFREQ_LIBS)
+
+if HAVE_POLKIT
+mate_cpufreq_applet_LDADD += $(POLKIT_LIBS)
```
muktupavels, line 44: This now adds `polkit-gobject-1` library here but polkit it self is not used in applet.

rbuj: You're right, here it just needs to use DBUS LIBS & CFLAGS. I also removed MATEDESKTOP LIBS & CFLAGS. Test: `git grep MATEDESKTOP` , `grep MATEDESKTOP config.log`.

rbuj: DBUS LIBS & CFLAGS are no longer required in this file after porting to GDBus the clients of cpufreq-selector.

cpufreq/src/cpufreq-selector.c
```diff
@@ -154,19 +169,24 @@ selector_set_frequency_async (SelectorAsyncData *data)
...
-	dbus_g_proxy_begin_call_with_timeout (proxy, "SetFrequency",
-					      dbus_set_call_notify_cb,
-					      data, NULL,
-					      INT_MAX,
-					      G_TYPE_UINT, data->cpu,
-					      G_TYPE_UINT, data->frequency,
-					      G_TYPE_INVALID,
-					      G_TYPE_INVALID);
+	if (!cpufreq_selector_create_proxy (data->selector, &error)) {
+		g_warning ("%s", error->message);
```
raveit65, line 173: Indents...

cpufreq/src/cpufreq-selector.c
```diff
@@ -205,19 +224,25 @@ selector_set_governor_async (SelectorAsyncData *data)
...
-	dbus_g_proxy_begin_call_with_timeout (proxy, "SetGovernor",
-					      dbus_set_call_notify_cb,
-					      data, NULL,
-					      INT_MAX,
-					      G_TYPE_UINT, data->cpu,
-					      G_TYPE_STRING, data->governor,
-					      G_TYPE_INVALID,
-					      G_TYPE_INVALID);
+	if (!cpufreq_selector_create_proxy (data->selector, &error)) {
+		g_warning ("%s", error->message);
```
raveit65, line 228: Indents.....

raveit65:
Thanks for taken all the suggestions in account :)
I will start testing.

cpufreq/src/cpufreq-selector.c
raveit65: Opps and sorry, forget what i said, i didn`t saw this brackets {}.

cpufreq/src/cpufreq-selector.c
raveit65: Same here, sorry...

raveit65:
Thanks, this is much better. In detail:
- Installation with --enable-polkit runs fine and changing frequency or govenors with a password query works like a charm.
- As expected building failed when dbus-glib isn't installed because of left overs in `cpufreq-selector-service.c/h`.
```
make[5]: Entering directory '/media/Data/Programme/Linux/Mate-Desktop/git-version/github-matedesktop/mate-applets-dist/cpufreq/src/cpufreq-selector'
  CC       cpufreq-selector-service.o
cpufreq-selector-service.c:21:10: fatal error: dbus/dbus-glib-lowlevel.h: No such file or directory
   21 | #include <dbus/dbus-glib-lowlevel.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
Currently, configure doesn't abort, not sure how this should be handle?
Porting this file to GDbus, or let configure stop when `dbus-glib` isn't installed and porting this file in another PR?

I am OK with both solutions.

rbuj:
configure aborts the configuration when the dbus-glib-devel package is not installed on the system and the polkit support has been enabled:
```
$ sudo dnf remove dbus-glib-devel && ./autogen.sh --prefix=/usr --enable-polkit
<cut>
checking for CPUFREQ_SELECTOR... yes
checking for DBUS... no
configure: error: PolicyKit support explicitly requested but dbus-glib-1 not found
```
Test on Fedora:
```
$ sudo dnf remove dbus-glib-devel
$ pkg-config --cflags --libs dbus-glib-1
Package dbus-glib-1 was not found in the pkg-config search path.
Perhaps you should add the directory containing `dbus-glib-1.pc'
to the PKG_CONFIG_PATH environment variable
Package 'dbus-glib-1', required by 'virtual:world', not found
```

raveit65:
Yes with using `./configure --enable-polkit` configure failed, but when using only `./configure` without any flag config sumary is:
```
- cpufreq                  yes
            - building selector    yes
            - using PolicyKit      yes
            - enabling suid bit    no
```
and compiler failed.
Any way, not a big issue.

rbuj:
@raveit65 I think it's fixed. Test:
```
$ sudo dnf remove -y dbus-glib-devel && ./autogen.sh --prefix=/usr
<cut>
checking for CPUFREQ_SELECTOR... yes
checking for DBUS... no
configure: error: PolicyKit support requested but dbus-glib-1 not found
```

raveit65:
Yeap, this works fine.

By the way, for fedora building in never used mate-desktop-libs as build requires, so i am still wondering why this was here in mate-applets?

Can someone please review the GDbus port, please?

raveit65:
GDBus port looks similar to gnome-panel code. So i think it is OK.

PR #459, opened 2020-05-27
geyes: Add themes.ui

rbuj:


raveit65:
The new preferences UI looks a bit squashed (vertical).
old:
geyes-preferences-old
new:
geyes-preferenences-new

I think the default-width/height is a bit unusual.
Bildschirmfoto zu 2020-06-21 13-24-46

And i am wondering why a GtkFrame is used between GtkBox and GtkAlignment ?
Bildschirmfoto zu 2020-06-21 13-28-41

raveit65:
In general i don't mind too because i don't use this applet and i don't know any user which use this old applet.
And i didn't ask for a change or want to block it. I only ask some question :)
It's Ok that scrollbar isn't visible for such a entries, but the 2:3 format looks real strange an is inconsistent with other dialogs.
Did you ever heard about the `golden ratio` ?
https://en.wikipedia.org/wiki/Golden_ratio

selectiveduplicate:
I get what you're saying about the ratio and consistency. Being a scrollable window, maybe remove the default parameters for h-w? 
I've heard about about golden ratio but don't know about its application in software/computing.
@rbuj Any thoughts?

raveit65:
@selectiveduplicate 
The old TV format 4:3 or old vga formats use the golden ratio. So all windows looks like this.

raveit65:
But i f you like the new format of the dialog than go ahead and mergeit.
As is said i don't block this.....

rbuj:
done

PR #466, opened 2020-06-08
charpic-about: fix DE name in about dialog

selectiveduplicate:


raveit65:
@selectiveduplicate 
I think it is OK when you commit such a simple change directly for master and latest stable branch (1.24).
If it's build Ok than it is fine, You can also check building and generating a tarball with `./autogen.sh && make && make distcheck`, to be sure.
Let me know when you need write access to a repo.

PR #467, opened 2020-06-14
multiload: improve the code used for graphics

rbuj:
Screenshot displaying Multiload Applet and MATE System Monitor application

Screenshot at 2020-06-14 08-59-39

### DIFF
--- a/multiload/global.h
+++ b/multiload/global.h
@@ -46,7 +46,7 @@ struct _LoadGraph {
     long cpu_last [NCPUSTATES];
     int cpu_initialized;
 
-    double loadavg1;
+    float percentage_used;
     NetSpeed *netspeed_in;
     NetSpeed *netspeed_out;
     guint net_threshold;
--- a/multiload/linux-proc.c
+++ b/multiload/linux-proc.c
@@ -250,26 +250,22 @@ GetPage (int Maximum, int data [3], LoadGraph *g)
 #endif /* 0 */
 
 void
-GetMemory (int Maximum, int data [5], LoadGraph *g)
+GetMemory (int Maximum, int data [2], LoadGraph *g)
 {
-    int user, shared, buffer, cached;
-
+    int user;
     glibtop_mem mem;
+    float ratio;
 
     glibtop_get_mem (&mem);
 
     g_return_if_fail ((mem.flags & needed_mem_flags) == needed_mem_flags);
 
-    user    = rint (Maximum * (float)mem.user / (float)mem.total);
-    shared  = rint (Maximum * (float)mem.shared / (float)mem.total);
-    buffer  = rint (Maximum * (float)mem.buffer / (float)mem.total);
-    cached = rint (Maximum * (float)mem.cached / (float)mem.total);
+    ratio = (float)mem.user  / (float)mem.total;
+    g->percentage_used = 100.0f * ratio;
+    user = rint (Maximum * ratio);
 
-    data [0] = user;
-    data [1] = shared;
-    data [2] = buffer;
-    data [3] = cached;
-    data [4] = Maximum-user-shared-buffer-cached;
+    data[0] = user;
+    data[1] = Maximum - user;
 }
 
 void
@@ -301,10 +297,9 @@ GetLoadAvg (int Maximum, int data [2], LoadGraph *g)
 
     g_return_if_fail ((loadavg.flags & needed_loadavg_flags) == needed_loadavg_flags);
 
-    /* g->loadavg1 represents %used */
-    g->loadavg1 = loadavg.loadavg[0];
+    g->percentage_used = loadavg.loadavg[0];
 
-    data [0] = rint ((float) Maximum * g->loadavg1);
+    data [0] = rint ((float) Maximum * g->percentage_used);
     data [1] = Maximum - data[0];
 }
 
--- a/multiload/linux-proc.c
+++ b/multiload/linux-proc.c
@@ -280,9 +280,14 @@ GetSwap (int Maximum, int data [2], LoadGraph *g)
 
     if (swap.total == 0) {
         used = 0;
+        g->percentage_used = 0;
     }
     else {
-        used = rint (Maximum * (float)swap.used / swap.total);
+        float ratio;
+
+        ratio = (float)swap.used / (float)swap.total;
+        g->percentage_used = 100.0f * ratio;
+        used = rint (Maximum * ratio);
     }
 
     data [0] = used;
--- a/multiload/linux-proc.h
+++ b/multiload/linux-proc.h
@@ -8,7 +8,7 @@ G_GNUC_INTERNAL void GetDiskLoad (int Maximum, int data [3], LoadGraph *g);
 #if 0
 G_GNUC_INTERNAL void GetPage (int Maximum, int data [3], LoadGraph *g);
 #endif /* 0 */
-G_GNUC_INTERNAL void GetMemory (int Maximum, int data [4], LoadGraph *g);
+G_GNUC_INTERNAL void GetMemory (int Maximum, int data [2], LoadGraph *g);
 G_GNUC_INTERNAL void GetSwap (int Maximum, int data [2], LoadGraph *g);
 G_GNUC_INTERNAL void GetLoadAvg (int Maximum, int data [2], LoadGraph *g);
 G_GNUC_INTERNAL void GetNet (int Maximum, int data [4], LoadGraph *g);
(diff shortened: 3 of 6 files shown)


dark-penguin:
Could we please not remove the very useful functionality of showing different "types" of memory in the multiload applet? There is nothing at all to be gained from this removal, the same could be achieved by setting the same color for all types. This is the reason I like this applet so much, and it is the most important metric I use. This change will make the multiload applet completely useless to me. If there is a reason we really need this removed, maybe have it optional, maybe even at compile time to make it simpler?

raveit65:
What is worst to display all mem types in the applet?

dark-penguin:
See #464 : a while ago the kernel changed how memory types are assigned, so the applet no longer shows them correctly. I've studied this, and it's now a little trickier - "shared" now means only tmpfs, but tmpfs is also counted as "used" and "cache", so you have to subtract it to show correctly.

I really needed this and nobody seemed to care, so I've learned C and fixed this myself. :) But the resolution is that "glibtop says nobody cares anymore, it's too complex now, just show MemAvailable as free and the rest is... not free".

rbuj:
> See #464 : a while ago the kernel changed how memory types are assigned, so the applet no longer shows them correctly. I've studied this, and it's now a little trickier - "shared" now means only tmpfs, but tmpfs is also counted as "used" and "cache", so you have to subtract it to show correctly.

Mem: shared (can be ignored nowadays. It has no meaning.)
https://access.redhat.com/solutions/406773

dark-penguin:
I don't know about Red Hat, but on my Debian and Ubuntu systems, "shared" is tmpfs. It is also counted as "used", so yes, it *can* be ignored, but it does not *have to be* ignored, and I find it very useful to see tmpfs separately. If someone wants to ignore it, they can set its color to the same as "Used".

Actually, you might do that by default; this way, it is simple for those who don't understand memory, but can be changed for those who want tmpfs separately.

Also, it is nice to see cached memory too.

https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt

rbuj:
> https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt
> (I don't remember where did I see that "shared" is only "tmpfs" now, but I've studied this issue for a long time. And anyway, all "shared" is counted as "used", so it's completely safe to subtract it from "used" and show separately.)

shared is Shmem, take a look at libgtop:
https://gitlab.gnome.org/GNOME/libgtop/blob/master/sysdeps/linux/mem.c#L59

Shmem is not only the memory used by tmpfs.
See https://www.kernel.org/doc/Documentation/filesystems/proc.txt
> Shmem: Total memory used by shared memory (shmem) and tmpfs

dark-penguin:
Shmem is "almost" only tmpfs - and some other memory that is also always non-available, so it's always part of "used" memory.

I also have a little more "shmem" than "tmpfs", however in your case I suspect that you forgot to unmount your first tmpfs (you put "mount" into your script, but no "unmount" anywhere, and your shmem is almost exactly twice the size of your test file.)

dark-penguin:
I've read through that part of glibtop code when trying to find a solution for this problem. I know how it parses /proc/meminfo and then gives you values from there. And then, it gives you something called "user" - meaning, "all memory that's not available". Simply using it is the easiest way, but we can subdivide both parts further.

"Used" consists of "shared" and "otherwise used", and "available" consists of "free" and "not free, but available". There is no damage in giving more information instead of less.

If you want to show a reliable difference between "used memory" and "available memory", then we can show all non-evictable cache as "used", like this:

- Free (`free`)
- Cache (meaning, available but not free; would be MemAvailable - MemFree, but the only way to get "MemAvailable" from glibtop is `total - user`, so `total - user - free`)
- Shared (`shared`)
- Used (`user - shared`)

Actually, I think this is better than my initial proposal. Buffers are irrelevant; what does matter is how much memory is used and how much is available, and then subdivide them more precisely: "used - as shared or otherwise?", and "available - but free or cached?".

rbuj:
You are right I called it twice, without "umount" :)
```
[robert@localhost ~]$ cat test2.sh 
echo "BEFORE"
free --kilo
cat /proc/meminfo | grep "^Shmem:.*"
df -t tmpfs -k | grep "^tmpfs" | awk '{total = total + $3}END{print total}'
echo "-----------"
sudo mount -t tmpfs -o size=1024m tmpfs /mnt/ramdisk
head -c 512MB /dev/zero > /mnt/ramdisk/file.txt
echo "AFTER"
free --kilo
cat /proc/meminfo | grep "^Shmem:.*"
df -t tmpfs -k | grep "^tmpfs" | awk '{total = total + $3}END{print total}'
[robert@localhost ~]$ ./test2.sh 
BEFORE
              total        used        free      shared  buff/cache   available
Mem:       33667661      331567    32742289        5193      593805    32931643
Swap:       6442446           0     6442446
Shmem:              5072 kB
1232
-----------
AFTER
              total        used        free      shared  buff/cache   available
Mem:       33667661      332746    32226500      517029     1108414    32417837
Swap:       6442446           0     6442446
Shmem:            504912 kB
501232
```
Red Hat documentation says that free's shared field can be ignored nowadays for RHEL 5, 6, 7.0, but it's Shmem /proc/meminfo in RHEL 7.1 or later (procps-ng 3.3.10) and glibtop for about four years.

The test also shows that Shmem is not equal to the sum of the space used in tmpfs file systems.

Memory used as of RHEL 7.1 is MemTotal - MemFree - Buffers - Cached - Slab
```
$ cat test3.sh 
declare -i used
declare -i not_available
declare -i MemTotal
declare -i MemFree
declare -i Buffers
declare -i Cached
declare -i Slab
declare -i MemAvailable
declare -i diff
while read -r line
do
  set -- $line
  case $1 in
    "MemTotal:")
    MemTotal=$2
    ;;
    "MemFree:")
    MemFree=$2
    ;;
    "Buffers:")
    Buffers=$2
    ;;
    "Cached:")
    Cached=$2
    ;;
    "Slab:")
    Slab=$2
    ;;
    "MemAvailable:")
    MemAvailable=$2
    ;;
  esac
done < /proc/meminfo

used=$MemTotal-$MemFree-$Buffers-$Cached-$Slab
echo "used = $used = $MemTotal - $MemFree - $Buffers - $Cached - $Slab"

not_available=$MemTotal-$MemAvailable
echo "not_available = $not_available = $MemTotal - $MemAvailable"

diff=$not_available-$used
echo "diff = $diff = $not_available - $used"

echo ""
echo "free --kilo"
free --kilo
```
Run
```
$ ./test3.sh 
used = 214084 = 32878576 - 31454516 - 57004 - 1000924 - 152048
not_available = 1207704 = 32878576 - 31670872
diff = 993620 = 1207704 - 214084

free --kilo
              total        used        free      shared  buff/cache   available
Mem:       33667661      318849    32209158      517132     1139654    32430706
Swap:       6442446           0     6442446
```
This calculation not accurate for displaying the memory pressure graph for several reasons and it's only an inaccurate estimation, this is one of the factors for which it was decided to introduce the MemAvailable field in /proc/meminfo

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

dark-penguin:
It is indeed quite complex to understand all the ways memory is used (I don't :) ). I agree now that having the graph based on MemAvailable is better than trying to properly display all kinds of memory. A simple "Used/Available" graph is indeed more informative in terms of "how much memory do I have left".

My argument is, there is no harm in displaying part of "available" memory as "really free", and part of "used" memory as "used as shared". Maybe even make them the same color by default. It is still the same graph based on MemAvailable, and "shared" and "free" are values you don't even need to calculate. But in addition, you can easily see how warmed-up your cache is, and easily track your tmpfs usage.

(I also have a little more shmem than tmpfs, but it is not significant. It is still definitely part of "used" memory in our MemAvailable-based graph, because it's not "available".)

raveit65:
> Memory used as of RHEL 7.1 is MemTotal - MemFree - Buffers - Cached - Slab

What is `Slab` ?
I don't understand why we shouldn't display `Buffers` and `Cached` any more?

rbuj:
> What is `Slab` ?

Slab: in-kernel data structures cache
https://www.kernel.org/doc/Documentation/filesystems/proc.txt

> I don't understand why we shouldn't display `Buffers` and `Cached` any more?

They are no longer part of the memory used, since part of the allocated memory can be claimed and therefore can be released, increasing the portion of free memory when this is required..

rbuj:
I close this PR to avoid duplicate reviews.

PR #486, opened 2020-07-08
battstat: warning about boolean algebra reported by cppcheck

rbuj:
```
$ cppcheck --enable=all . 2> err.txt
$ grep knownConditionTrueFalse err.txt 
battstat/acpi-freebsd.c:200:12: style: Condition 'charging' is always true [knownConditionTrueFalse]
```

### DIFF
--- a/battstat/acpi-freebsd.c
+++ b/battstat/acpi-freebsd.c
@@ -197,7 +197,7 @@ acpi_freebsd_read(struct apm_info *apminfo, struct acpi_info * acpiinfo)
   if (!charging) {
     apminfo->ai_batt_time = time;
   }
-  else if (charging && rate > 0) {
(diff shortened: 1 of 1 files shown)


raveit65:
What is the content of the warning?
Looks like you change the condition, or not?
```
-  else if (charging && rate > 0) {
+  else if (rate > 0) {
```
No idea how to test this without freebsd.

rbuj:
The else block (2) implies that the condition of the if statement (1) is false, then the condition (2) can be simplified from (!A && B) to (B):
```
if (A) { /* A (1) */
 /* code */
} else if (!A && B) { /* !A, B (2) */
 /* code */
} else { /* !A, !B (3) */
 /* code */
}
```

PR #487, opened 2020-07-14
battstat: fix deprecated up_client_get_devices

raveit65:
The previous code would have leaked all the UpDevice objects because no
free function was set on the returned GPtrArray.

With depending on upower-glib 0.99.8 get_devices() was deprecated and get_devices2() was introduced, we
can simply switch to get_devices2() which does set a free function on
the returned GPtrArray, stopping the leak.


rbuj:
I think that we can also bump upower to 0.99.8 and remove legacy code.

https://github.com/mate-desktop/mate-applets/blob/01e8f8a09ab273719bdf0869f9f8693cb45d302e/configure.ac#L21

raveit65:
You're right, done.

raveit65:
Origin comment was taken from https://github.com/mate-desktop/mate-power-manager/commit/9d5b257f74d396ad4ed3d144185ab95b318b03b8
which was from 
https://gitlab.gnome.org/GNOME/gnome-control-center/commit/c1210c5
https://gitlab.freedesktop.org/upower/upower/commit/cb1071b

I think it is more a leak in upower itself.

rbuj:
Using a global variable makes more sense like in #443, since the list of devices is updated dynamically and specific events are triggered.

raveit65:
I only wanted to fix the deprecation. Any improvements can be done in another PR.

PR #490, opened 2020-07-15
multiload: Fix memory leak

rbuj:


rbuj:
The merge conflict has been fixed.

PR #488, opened 2020-07-14
multiload: Display the percentage used accurately on tooltip

rbuj:

PR #493, opened 2020-07-16
multiload: memload graph on Linux

rbuj:
Closes #464

### Screenshots

Screenshot at 2020-07-17 00-34-34

Screenshot at 2020-07-16 22-28-10

### DIFF
--- a/multiload/global.h
+++ b/multiload/global.h
@@ -35,7 +35,9 @@ typedef enum {
 
 typedef enum {
     memload_user = 0,
+#ifndef __linux__
     memload_shared,
+#endif /* __linux__ */
     memload_buffer,
     memload_cached,
     memload_free,
--- a/multiload/linux-proc.c
+++ b/multiload/linux-proc.c
@@ -256,6 +256,18 @@ GetPage (int Maximum, int data [3], LoadGraph *g)
 }
 #endif /* 0 */
 
+/* GNU/Linux: The Shared memory (mem.shared) is part of the Cached memory
+ * (mem.cached).
+ *   aux [memload_user]   = (mem.total - mem.free) - (mem.cached + mem.buffer)
+ *   aux [memload_buffer] = mem.buffer;
+ *   aux [memload_cached] = mem.cached;
+ *
+ * Other operating systems:
+ *   aux [memload_user]   = mem.user;
+ *   aux [memload_shared] = mem.shared;
+ *   aux [memload_buffer] = mem.buffer;
+ *   aux [memload_cached] = mem.cached;
+ */
 void
 GetMemory (int        Maximum,
            int        data [memload_n],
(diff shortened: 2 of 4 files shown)


dark-penguin:
This introduces a lot of platform-specific code just to avoid displaying shared memory, which is all part of cache, so it can be safely displayed. In most cases, it would be less than one pixel in size, but if this seems confusing, then it could have the same default color as cache, and behave exactly as it is here.

Also, this displays useless metrics like "cache" and "buffers", but does not display the most important metric - "how much available memory do I have".

rbuj:
It needs to be reworked after #488

raveit65:
@rbuj 
So, this one next?
https://github.com/mate-desktop/mate-applets/pull/488 is in master.

dark-penguin:
@raveit65 Are you sure you want to add this much platform-specific code just to avoid displaying the shared part of cache separately? Maybe even with the same default color, if that bothers you. Maybe even just leave it in and pass zero as its value; I could fix the rest myself in a local patch, but this is way too much to revert. Not to mention someone is probably going to touch some of this code again, so I can't even keep a patch that will work.

raveit65:
What code is platform-specific?
Which distro won't support this code?
What mem is missing for you?
Honestly, i would like to see an agreement between you two.
I don't like to decide this, ....and i can wait :)

dark-penguin:
Since we are not in agreement, I'm trying to get other members to evaluate this and make a decision. I would be ok with a decision not in my favor, as long as it is really a decision made by several members, and not one person who I think is trying to do the wrong thing, which could be corrected if somebody else noticed.

> What code is platform-specific?
> Which distro won't support this code?
> What mem is missing for you?

Take a look at the code in this PR. It removes "shared" memory for no good reason (the only explanation I got is that "it would create confusion"), and to do that only on Linux, it goes to great lengths to check *in multiple places* which platform are we running, and does things differently if we are on Linux. This is much worse than just passing a zero in one place as its value. Which is not even necessary, because we could just display it.

This 300+ changed lines PR is far worse than what I did with three changed lines. Not only it drops functionality AND backwards compatibility, but it introduces a lot of unnecessary complexity *in order to drop it*.

Is it just me, or does this look like a very bad solution? @lukefromdc , this is what it would look like if we wanted to remove that color instead of passing that zero. :) Which is why I immediately suggested to not even think about that. It does not even "fix" the way colors are stored to prevent this problem in the future; instead, it introduces a specific fix somewhere else for a specific color in a specific graph!

Could someone please evaluate which one does the job better, this or #464 ? In https://github.com/mate-desktop/mate-applets/pull/464#issuecomment-659627420 , I have explained in great detail why do I think we can display all memory categories the way I did it. I've tried to get any meaningful arguments against my solution, but did not get much, other than "just do this or go away".

But even if you do not agree about that, just please find a way to avoid this much Linux-specific code! This will be very hard to revert and maintain!

rbuj:
@raveit65 #488 is the next

rbuj:
This pull request need to be reworked after #488

rbuj:
It's ready. In case you want to continue showing the value of the shared memory, merge #513 and close this PR.

PR #501, opened 2020-07-20
geyes: Load menu from resources

rbuj:


lukefromdc:
Merging as this is unlikely to get another review

PR #502, opened 2020-07-20
multiload: Remove legacy variable 'show_frame'

rbuj:


raveit65 (approved):
LGTM,
builds and runs like before.

PR #510, opened 2020-07-29
Netspeed kernel warning

rbuj:


raveit65:
Kernel warnings are gone and connecting to wifi-network works like before.
```
[root@mother rave]# dmesg | grep netlink
[    0.432920] audit: initializing netlink subsys (disabled)
[    1.190899] Initializing XFRM netlink socket
```
Thanks.

PR #514, opened 2020-07-30
multiload: Do not get the value of the "diskload-nvme-diskstats" on all calls to GetDiskLoad

rbuj:


raveit65:
For some reasons i need to restart the applet to see the load of my nvme-disks after i enabled the checkbox in preferences.
But master has same behaviour, so it's not related to PR.

raveit65 (approved):
LGTM

PR #515, opened 2020-07-30
multiload: do not use a computed compound key for graph visibility

rbuj:


raveit65 (approved):
LGTM,
builds and runs fine.

PR #518, opened 2020-07-30
cpufreq: Update cpufreq-preferences.ui

rbuj:


raveit65:
Can we move the help button to the left please, like it was before?
cpufreq-prefs
And `show frequency units ` should be indent.

rbuj:
@raveit65 Sometimes glade does not render the controls correctly.

Screenshot at 2020-08-03 15-37-22

raveit65:
Ok, i will install it :)

PR #525, opened 2020-08-07
battstat: use gtk_image_set_from_icon_name for setting status icon

rbuj:


raveit65:
Hmm, without testing, does this remove HIDPI scaling support?

rbuj:
No, icons are already scaled. I don't detect any HiDPI issue.

Screenshot at 2020-08-07 17-58-13

Screenshot at 2020-08-07 17-59-18

raveit65:
Main different is that the icon doesn't change size any more, but problem exists.
The icon doesn't fit in panel with default size of 24 px.
If i increase the panel size to 27px, than the icon displays.
And now i got this issue with my main box with HIDPI monitor too.
Still wondering that you didn't see that with your system?

PR #541, opened 2020-11-22
build: replace acinclude.m4 with original as-ac-expand.m4

rbuj:


raveit65 (approved):
Builds fine at travis.

PR #547, opened 2020-11-28
build: Do not define M4_FILES twice

rbuj:


raveit65 (approved):
LGTM,
it builds fine at travis.

PR #550, opened 2020-12-02
mateweather: Use common subdirs - src, data

rbuj:


rbuj:
It only fixes a merge conflict.

raveit65 (approved):
Builds and runs like before.

PR #551, opened 2020-12-02
drivemount: Use common subdirs - data, src

rbuj:


rbuj:
It only fixes a merge conflict.

vkareh (approved):
LGTM

I see no build issues and the applet works same as before

raveit65 (approved):
Builds and runs fine, same as before.

PR #552, opened 2020-12-02
stickynotes: Add stickynotes-applet-menu.xml into resource bundle

rbuj:


vkareh (approved):
LGTM
Works the same for me

raveit65 (approved):
LGTM,
builds and runs fine, same like before.

PR #553, opened 2020-12-02
trashapplet: load menu and ui files from resource bundle

rbuj:


raveit65 (approved):
LGTM,
builds and runs fine.

PR #555, opened 2020-12-04
remove dead function mate_panel_applet_set_background_widget

mbkma:


rbuj (approved):
It removes the calls to the empty function mate_panel_applet_set_background_widget.
It also removes trailing whitespaces.

PR #558, opened 2020-12-10
netspeed: show ipv6 address in tooltip if ipv4 address is not used

### LINKED ISSUE #557: netspeed: support IPv6 in tooltip window
#### Expected behaviour

When hovering the applet in the status bar, it should show the IPv6 address of the NIC in the tooltip window.

#### Actual behaviour

When hovering the applet in the status bar, it only shows the legacy IPv4 address. If no legacy address is configured on that interface, it will only show "0.0.0.0" in the tooltip window.

Configure interface with IPv6 address. Add netspeed applet to taskbar. Hover it with the mouse for the tooltip window to appear.

#### MATE general version

MATE Desktop Environment 1.24.0

#### Package version

MATE Netspeed 1.24.0

#### Linux Distribution

Mint 20 MATE 64-bit

rbuj:
Closes #557

### DIFF
--- a/netspeed/src/netspeed.c
+++ b/netspeed/src/netspeed.c
@@ -873,9 +873,9 @@ update_applet (MateNetspeedApplet *applet)
 #endif
     }
 
-    update_tooltip(applet);
+    update_tooltip (applet);
 
-    /* Refresh the text of the labels and tooltip */
+    /* Refresh the text of the labels */
     if (applet->show_sum) {
         gtk_label_set_text (GTK_LABEL (applet->sum_label),
                             applet->devinfo->sum_rate);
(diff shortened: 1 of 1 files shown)


vkareh (approved):
LGTM
<del>I don't have ip6-only interfaces, but code looks fine and applet builds and runs well for me</del>

_edit: I have an ipv6 `vnet0` device and now it actually shows the IPv6 address with this patch_
:+1:

raveit65 (approved):
LGTM,
i don't have an ipv6 only device, but applet runs fine.

treysis:
Thanks for fixing this so fast!
One addition though: I think it should show the IPv6 regardless of IPv4 being used or not.

rbuj:
[IPv4 addressing](https://en.wikipedia.org/wiki/IPv4) is valid for [LANs](https://en.wikipedia.org/wiki/Local_area_network) and it's commonly used for [private networks](https://en.wikipedia.org/wiki/Private_network), since [NAT](https://en.wikipedia.org/wiki/Network_address_translation) fixes the [IPv4 address exhaustion](https://en.wikipedia.org/wiki/IPv4_address_exhaustion) issue.

treysis:
@rbuj Unless you are Microsoft or Facebook...then even NAT isn't enough. And no, some people run IPv6-only networks. Some mobile networks have long gone IPv6-only as well, so this applies also to tethered devices (e.g. your Linux Mint laptop tethered to your phone!). And IPv4 is legacy, IPv6 is the future. Linux always being spoken of as being great with networks should especially make use of IPv6 wherever possible.

treysis:
I'll continue discussion in the issue https://github.com/mate-desktop/mate-applets/pull/558#issuecomment-744356830

PR #559, opened 2020-12-10
multiload: increase refresh rate range setting

### LINKED ISSUE #432: multiload applet "Update interval" limited to 10 sec max
Any particular reason why the applet update interval can't be set more than 10 seconds  ?

I often have long-running process, especially at night, and it is convenient for me to get a quick overview of the past activity during, say, the last 12 hours period.

It looks it is only a simple UI issue:
at multiload/properties.c : (l590):
> 	spin_button = gtk_spin_button_new_with_range(50, 10000, 50);

Raising the maximum to, say, five minutes, would be very helpful to fit the past 12h activity on a reasonably-sized panel applet.

rbuj:
Closes #432

### DIFF
--- a/multiload/src/load-graph.c
+++ b/multiload/src/load-graph.c
@@ -265,8 +265,9 @@ load_graph_unalloc (LoadGraph *g)
     g->pos = NULL;
     g->data = NULL;
 
-    g->size = g_settings_get_int(g->multiload->settings, "size");
-    g->size = MAX (g->size, 10);
+    g->size = CLAMP (g_settings_get_uint (g->multiload->settings, GRAPH_SIZE_KEY),
+                     GRAPH_SIZE_MIN,
+                     GRAPH_SIZE_MAX);
 
     if (g->surface) {
         cairo_surface_destroy (g->surface);
@@ -412,7 +413,7 @@ load_graph_new (MultiloadApplet *ma, guint n, const gchar *label,
     g->n = n;
     g->id = id;
     g->speed = speed;
-    g->size = MAX (size, 10);
+    g->size = size;
     g->pixel_size = mate_panel_applet_get_size (ma->applet);
     g->tooltip_update = FALSE;
     g->multiload = ma;
--- a/multiload/src/load-graph.c
+++ b/multiload/src/load-graph.c
@@ -411,8 +411,8 @@ load_graph_new (MultiloadApplet *ma, guint n, const gchar *label,
     g->name = name;
     g->n = n;
     g->id = id;
-    g->speed  = MAX (speed, 50);
-    g->size   = MAX (size, 10);
+    g->speed = speed;
+    g->size = MAX (size, 10);
     g->pixel_size = mate_panel_applet_get_size (ma->applet);
     g->tooltip_update = FALSE;
     g->multiload = ma;
--- a/multiload/src/load-graph.c
+++ b/multiload/src/load-graph.c
@@ -96,17 +96,13 @@ load_graph_draw (LoadGraph *g)
   {
     guint maxnet = 1;
     gint segments = 1;
-    gint combined;
     guint net_threshold;
 
     for (i = 0; i < g->draw_width; i++)
     {
       g->pos [i] = g->draw_height - 1;
-      combined = g->data[i][0] +
-                 g->data[i][1] +
-                 g->data[i][2];
-      if (combined > maxnet)
-        maxnet = combined;
+      if (g->data[i][3] > maxnet)
+        maxnet = g->data[i][3];
     }
     //printf("max = %d ", maxnet);
     guint level = 0;
--- a/multiload/src/load-graph.c
+++ b/multiload/src/load-graph.c
@@ -30,15 +30,15 @@
 static void
 shift_right(LoadGraph *g)
 {
-    unsigned i;
-    int* last_data;
+    guint64 *last_data;
+    gsize i;
 
     /* data[g->draw_width - 1] becomes data[0] */
     last_data = g->data[g->draw_width - 1];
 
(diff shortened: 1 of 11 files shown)


raveit65:
Hmm, here the spinbutton for update-interval adjust the height of the graph :open_mouth: 
Looks like that both spinbuttons are reversed in preferences-dialog.

rbuj:
@raveit65 done

raveit65:
Ok,  that's better. But can we reduce the stepping for both spinbuttons if possible please?
- spinbutton for height of the graph use a 5px stepping, better 1px or not?
- spinbutton for update-interval use a stepping of 50ms, better 10ms?

For me the best range is between 50 - 100ms to see a fluid nice graph.

raveit65:
Something is wrong with showing the output network graph. (green ,third graph from top, sending ~500kB/s)
With master:
network-graph-normal

With PR the green graph for sending is to small (nearly zero):
network-graph-wrong

I think incoming graph is to small too.

rbuj:
I checked the network graph traffic and it's displayed like before.
The area of the graph may be different, which depends on the previous traffic.
IMHO, the default values are fine, increasing the sampling rate would increase the cpu load.
I completely agree with you, the value of all steps should be updated, it can be easily done in the user interface file. maybe it would be better to do it in another PR for not overloading this one.

### Outgoing traffic
out

### Incoming traffic
in

raveit65:
@rbuj 
In your sreenshot is the graph of mate-netspeed-applet which works fine.
But i meant the graph of multiload-applet in panel.
network-graph-wrong
Here the upload  (green) is to small.

raveit65:
Btw. you need to change the default color for uploads in network graph of multiload-applet.
I think default is to use yellow color for upload and download, which is weird.

rbuj:
Test on Fedora using iperf:

HOST A - listen 5001/udp
```
$ sudo firewall-cmd --zone=public --add-port=5001/udp
$ iperf -s -u
```

HOST B - send 5001/udp 10mbps for 60 seconds
```
$ iperf -c 192.168.1.9 -u -b 10m -t 60
```

rbuj:
There is an issue showing the outbound traffic, but it's in the master branch too.

Screenshot at 2020-12-14 23-12-53

Test: IN 10mbps, OUT 20mbps

raveit65:
With master branch it works fine for me, only with PR the graph shows nearly zero output when i upload  ~500kB/s with transmission-client.

rbuj:
> With master branch it works fine for me, only with PR the graph shows nearly zero output when i upload ~500kB/s with transmission-client.

I think the data sources are stacked on the network graph.

multiload/src/load-graph.c
```diff
@@ -96,17 +96,13 @@ load_graph_draw (LoadGraph *g)
...
     {
       g->pos [i] = g->draw_height - 1;
-      combined = g->data[i][0] +
-                 g->data[i][1] +
-                 g->data[i][2];
-      if (combined > maxnet)
-        maxnet = combined;
+      if (g->data[i][3] > maxnet)
+        maxnet = g->data[i][3];
     }
```
raveit65, line 106: Build warning.
```
load-graph.c: In function 'load_graph_draw':

load-graph.c:104:25: warning: comparison of integer expressions of different signedness: 'gint' {aka 'int'} and 'guint' {aka 'unsigned int'} [-Wsign-compare]

  104 |       if (g->data[i][3] > maxnet)
```
https://travis-ci.org/github/mate-desktop/mate-applets/jobs/749808750#L12206
Possible it was already there.

raveit65:
This works much better now, good work.

> I completely agree with you, the value of all steps should be updated, it can be easily done in the user interface file. maybe it would be better to do it in another PR for not overloading this one.

+1

multiload/src/load-graph.c
rbuj: Yes, you're right. There are a lot of sign and width conversion errors.

libgtop mainly uses guint64 datatype for data, we use int for graph data, and cairo uses double datatype. And there are some unsigned longs in the middle... can use guint64 for graph data, although it may require a bit more memory.
```C
typedef void (*LoadGraphDataFunc) (int, int [], LoadGraph *);
```
```
multiload/src/load-graph.c:250:load_graph_unalloc (LoadGraph *g)
multiload/src/load-graph.c:280:load_graph_alloc (LoadGraph *g)
```

I also think some variables should be promoted to `gsize`, if they store a sizeof value or they are used for a malloc later.

raveit65: Sounds like another PR to fix it?

rbuj: Ok, I'm going to fix `guint net_threshold` and `*threshold*` in this PR, which is stored using 32-bit unsigned integers in gsettings and then stored in this variable as 16-bit unsigned integer. We need at least 30-bit uint = (log 1000000000 / (log 2) = 30
```
    <key name="netthreshold3" type="u">
      <range min="12" max="1000000000"/>
      <default>100000000</default>
      <summary>Network threshold 3 in bytes</summary>
    </key>
```

rbuj: @raveit65 done

rbuj: It's not ready to go, I detected an error in the network graph...

rbuj: @raveit65 done

raveit65: Warning is gone. thx.
What a huge commit...

raveit65 (approved):
LGTM,
builds aund runs fine.
Network graph displays fine now.

vkareh (approved):
This works well for me as well. Code change is large, but looking at the individual commits and it looks good :+1:

PR #560, opened 2020-12-12
mateweather: Avoid NULL pointer dereferences if there is no weather info

rbuj:
```
$ LANG=C journalctl --no-pager --boot | grep mateweather
Dec 11 21:32:06 localhost.localdomain mateweather-app[1617]: weather_info_get_icon_name: assertion 'info != NULL' failed
Dec 11 21:32:06 localhost.localdomain mateweather-app[1617]: weather_info_get_temp_summary: assertion 'info != NULL' failed
Dec 11 21:32:06 localhost.localdomain mateweather-app[1617]: weather_info_get_icon_name: assertion 'info != NULL' failed
```

### DIFF
--- a/mateweather/src/mateweather-applet.c
+++ b/mateweather/src/mateweather-applet.c
@@ -122,14 +122,15 @@ static const GtkActionEntry weather_applet_menu_actions [] = {
 	  G_CALLBACK (about_cb) }
 };
 
-static void place_widgets (MateWeatherApplet *gw_applet)
+static void
+place_widgets (MateWeatherApplet *gw_applet)
 {
     GtkRequisition req;
     int total_size = 0;
     gboolean horizontal = FALSE;
     int panel_size = gw_applet->size;
-    const gchar *temp;
-    const gchar *icon_name;
+    const gchar *temp = NULL;
+    const gchar *icon_name = NULL;
 
     switch (gw_applet->orient) {
 	case MATE_PANEL_APPLET_ORIENT_LEFT:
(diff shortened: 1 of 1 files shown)


mateweather/src/mateweather-applet.c
```diff
@@ -143,25 +144,23 @@ static void place_widgets (MateWeatherApplet *gw_applet)
...
-    gw_applet->label = gtk_label_new("0\302\260F");
-
     /* Update temperature text */
-    temp = weather_info_get_temp_summary(gw_applet->mateweather_info);
-    if (temp)
-        gtk_label_set_text(GTK_LABEL(gw_applet->label), temp);
+    if ((gw_applet->mateweather_info != NULL) && ((temp = weather_info_get_temp_summary (gw_applet->mateweather_info)) != NULL))
+        gw_applet->label = gtk_label_new (temp);
+    else
+        gw_applet->label = gtk_label_new(_("?ºF"));
```
vkareh, line 163: Can we drop the `ºF` from here? If I have my temp in Celsius, when re-loading the applet in the panel there's a brief period in which it shows `?ºF`

Also, note the screenshot: it shows as a smaller text than the normal temp should be, and the degrees sign has a line under it. Perhaps we should revert it back to using `gtk_label_set_text` and continue using the unicode character codes?
Screenshot at 2020-12-14 09-53-58
Screenshot at 2020-12-14 09-55-22

rbuj: @vkareh done

rbuj: At first glance, I think the text size is smaller on your screenshot because the applet uses two rows for displaying the weather info. Perhaps we can display some icon when it cannot be got from the current weather info.

rbuj: done
```diff
diff --git a/mateweather/src/mateweather-applet.c b/mateweather/src/mateweather-applet.c
index 2048bfc8..ca07d686 100644
--- a/mateweather/src/mateweather-applet.c
+++ b/mateweather/src/mateweather-applet.c
@@ -146,15 +146,15 @@ place_widgets (MateWeatherApplet *gw_applet)
     /* Create the weather icon */
     if ((gw_applet->mateweather_info) && ((icon_name = weather_info_get_icon_name (gw_applet->mateweather_info)) != NULL)) {
         gw_applet->image = gtk_image_new_from_icon_name (icon_name, GTK_ICON_SIZE_BUTTON);
-        gtk_widget_show (gw_applet->image);
-        gtk_widget_get_preferred_size (gw_applet->image, &req, NULL);
-        if (horizontal)
-            total_size += req.height;
-        else
-            total_size += req.width;
     } else {
-        gw_applet->image = gtk_image_new ();
+        gw_applet->image = gtk_image_new_from_icon_name ("weather-storm", GTK_ICON_SIZE_BUTTON);;
     }
+    gtk_widget_show (gw_applet->image);
+    gtk_widget_get_preferred_size (gw_applet->image, &req, NULL);
+    if (horizontal)
+        total_size += req.height;
+    else
+        total_size += req.width;
 
     /* Update temperature text */
     if ((gw_applet->mateweather_info != NULL) && ((temp = weather_info_get_temp_summary (gw_applet->mateweather_info)) != NULL))
```

vkareh (approved):
Works a lot better now, thanks!

Not sure if `weather-storm` is the best option (or if we should just not show any icon?), but that's just nitpicking, I don't have a strong opinion on that.

raveit65 (approved):
Does what it promise.

PR #562, opened 2020-12-13
multiload: load menu from resource bundle

rbuj:


raveit65 (approved):
Does what it promise.

PR #564, opened 2020-12-14
netspeed: show all IP addresses on tooltip if enabled

rbuj:


raveit65:
@rbuj 
This branch needs to updated to solve the merge conflict.

rbuj:
@raveit65 done

The merge conflict has been resolved. I removed the commits about showing the prefix and the scope of the IPv6 address. Later, I'll take a look at the code for showing all the IPv6 addresses in the device details dialog. I think we only need to pull them when it updates the dialog data, i.e. there is no need to capture all IPv6 addresses when we take the data for the graph, for example when the interface has been changed in the preferences dialog.

raveit65 (approved):
LGTM,
the new gsettings toogle enables ipv6 address in tooltip nicely.
Tested with vpn-provider connection.

vkareh (approved):
LGTM

PR #566, opened 2020-12-17
multiload: Add properties.ui

rbuj:


multiload/data/properties.ui
```diff
@@ -0,0 +1,1424 @@
...
+  <object class="GtkImage" id="close_button_img">
+    <property name="visible">True</property>
+    <property name="can-focus">False</property>
+    <property name="icon-name">window-close</property>
+  </object>
+  <object class="GtkAdjustment" id="graph_size_spin_button_adj">
+    <property name="lower">10</property>
+    <property name="upper">1000</property>
+    <property name="value">40</property>
+    <property name="step-increment">1</property>
```
raveit65, line 14: +1

multiload/data/properties.ui
```diff
@@ -0,0 +1,1424 @@
...
+    <property name="lower">12</property>
+    <property name="upper">1000000000</property>
+    <property name="step-increment">500</property>
+    <property name="page-increment">1000</property>
+  </object>
+  <object class="GtkAdjustment" id="speed_spin_button_adj">
+    <property name="lower">50</property>
+    <property name="upper">60000</property>
+    <property name="value">500</property>
+    <property name="step-increment">50</property>
```
raveit65, line 45: Can we use `name="step-increment">10<` please?
My favorite value seems to be 120 :)

raveit65:
I think the colorbuttons inside the `gtk_grid` of the `gtk_notebook` are looking better when you use `columm-spacing=60` for all tabs.
And for the proc-, ram-, ant network-tab i would use `Gtk_Align_Center`.
Than it looks more like before. You can test this and play around with gtk3-inspector easily.
Before:
multiload-pref-before
With PR:
Bildschirmfoto zu 2020-12-18 18-40-43

But i don't have a strong opinion on using `Gtk_Align_Center`.

rbuj:
> I think the colorbuttons inside the `gtk_grid` of the `gtk_notebook` are looking better when you use `columm-spacing=60` for all tabs.

homogeneous distribution of space between the columns, the color button is centered on the column

raveit65:
> > I think the colorbuttons inside the `gtk_grid` of the `gtk_notebook` are looking better when you use `columm-spacing=60` for all tabs.
> 
> homogeneous distribution of space between the columns, the color button is centered on the column

Of course....
With `columm-spacing=12` the buttons are looking very squashed. There is a lot of space  in the notebook tab which you don't use , and it isn't like before.
Anyway, when other member of the team prefer look&feel from PR, i won't block it.

@mate-desktop/core-team 
Thoughts?

raveit65 (approved):
LGTM,
i like the new design of the colorbutton and that they all have the same width.

lukefromdc (approved):
This builds and runs fine, didn't see anything but the usual deprecation warnings

PR #571, opened 2020-12-21
trashapplet: Upgrade trashapplet-empty-progress.ui

rbuj:


lukefromdc (approved):
Builds and installs fine, empty-trash dialog from the panel applet seems unchanged when the trash icon in a restarted mate-panel is right-clicked and empty trash selected.

raveit65 (approved):
LGTM

PR #572, opened 2020-12-22
netspeed: improve display of IPv6 addresses in device details.

treysis:
Device-details

I think IPv6 address box should go directly below IPv4 as IPv6 is the new standard.

Also, I think it looks better and more meaningful if address scope is added to the back of the address in brackets, not in front. This is also more consistent with display of IPv4 addresses.


rbuj (approved):
It looks good to me. Thank you very much!

PR #574, opened 2020-12-22
netspeed: show IPv4 subnet length instead of mask

rbuj:
Screenshot at 2020-12-22 07-41-27
Screenshot at 2020-12-22 07-43-57

### DIFF
--- a/netspeed/data/netspeed-details.ui
+++ b/netspeed/data/netspeed-details.ui
@@ -214,34 +214,6 @@
                         <property name="position">1</property>
                       </packing>
                     </child>
-                    <child>
-                      <object class="GtkLabel" id="netmask_label">
-                        <property name="visible">True</property>
-                        <property name="can-focus">False</property>
-                        <property name="halign">start</property>
-                        <property name="label" translatable="yes">Netmask:</property>
-                        <property name="xalign">0</property>
-                      </object>
-                      <packing>
-                        <property name="expand">False</property>
-                        <property name="fill">True</property>
-                        <property name="position">2</property>
-                      </packing>
-                    </child>
-                    <child>
-                      <object class="GtkLabel" id="netmask_text">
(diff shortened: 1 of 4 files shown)


treysis:
I think netmask is more descriptive to most people than subnet length.

vkareh:
This builds and works well.

I think using CIDR notation is fine, but it might confuse some users, since IP addresses already have a subnet length of `/32`. Also this makes it so that you need to do mental math to understand the IP range. Maybe if you leave the CIDR notation, but then show routing prefix and/or mask in a tooltip (or maybe even the actual IP range), it could help clear confusion.

What do you think?

rbuj:
All IPv4 address have 32 bits. The left n bits are the network address (prefix) and the right 32 - n are the host address (suffix). There is no difference on computing the value of network and broadcast addresses from CIDR notation since you get the netmask from prefix.

vkareh:
Correct, I'm referring to the mental math.

Take for example:
* IP: 192.168.2.32
* Netmask: 255.255.255.0
I can easily, at a quick glance, know that my address range is `192.168.2.[0-255]`

Similar with
* IP: 192.168.2.32
* Routing prefix: 192.168.2.0/24
I can easily tell what my range is as well.

Now with
* 192.168.2.32/24
I need to recall what the `/24` stands for. Is it for `x.x.x.0`, or is it `x.x.0.0`, or even `x.x.[128-255].[0-255]`?
I need to either do the mental math to calculate the range of `/24`, or memorize the range tables for the different subnet lengths.

So the information is the same, of course, but it's too dense to be useful at a quick glance. Does that make sense?

raveit65:
I think we shouldn't make it complicated as more as possible.
A netmask is understandable by nearly everyone who use a device with network connection..
I am in doubt that most people know what a subnet length means.

rbuj:
Ok.

PR #576, opened 2020-12-22
WIP netspeed: remove ipv4 routing check

rbuj:


vkareh:
What's the purpose of this? With your branch I lose a lot of information from my netspeed window:

* `master`:
Screenshot at 2020-12-22 08-44-13

* `pull/576`
Screenshot at 2020-12-22 08-44-43

rbuj:
The `get_default_route` was added for selecting an interface with a gateway on startup. It's part of the legacy and It makes a lot of unnecessary calls to pull current device info. The network graph data is collected from network status snapshots, and these are taken periodically, so there is no need to add extra snapshot on each iteration. I'll check it on baremetal for testing a wireless connection.

rbuj:
@vkareh It worked fine on baremetal:

Screenshot at 2020-12-22 20-51-13

vkareh:
My tests are all in baremetal. This still doesn't work for my default/wireless device.

vkareh:
wait... poking around at it, the wireless device works, but it's no longer the default. For some reason your patch changes what my default interface is... ???

rbuj:
Ok, the code is still used for setting the Default interface name. However, it only takes into account the ipv4 routing table.

PR #584, opened 2020-12-27
Fix dependency check of autogen.sh

### LINKED ISSUE #583: autogen.sh doesn't check for autopoint-dependency
#### Expected behaviour

Should notify user of missing dependency `autopoint`.

#### Actual behaviour

`./autogen.sh` fails with error without meaningful message.

Run `./autogen.sh` without having `autopoint` installed.

#### MATE general version

Current development snapshot.

treysis:
Fixes #583

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -22,6 +22,11 @@ which yelp-build || {
     exit 1
 }
 
+which autopoint || {
+    echo "You need to install autopoint"
+    exit 1
+}
+
 REQUIRED_AUTOMAKE_VERSION=1.9
 USE_COMMON_DOC_BUILD=yes
 


rbuj:
Internationalization of packages is optional, so autopoint can't be mandatory.
https://www.gnu.org/software/gettext/manual/html_node/lib_002fgettext_002eh.html

rbuj:
`autopoint` is already checked on `autoreconf ` sentence in `mate-autogen` file

treysis:
But if you just compile mate-applets, shouldn't there be a separate check? Not everyone wants to compile the whole mate desktop suite.

rbuj:
Take a look at the last line of the autogen.sh file. 
```
. mate-autogen
```

treysis:
It fails before that.

rbuj:
> It fails before that.

It's not true. Test: adding -xv in shebang.

PR #590, opened 2021-01-04
netspeed: allow to select an interface which is down

rbuj:


raveit65 (approved):
LGTM,
now i can select all devices and nothing crashes with all the other PRs.
But i have no idea when and what you want to get merged:)

PR #592, opened 2021-01-05
netspeed: avoid a NULL pointer dereference

rbuj:


raveit65 (approved):
This fixes the crash of the applet when re-opening preferences dialog after selecting selecting another device.
I can select all active devices now.
LGTM

raveit65:
Should i merge this?

PR #598, opened 2021-01-08
charpick: Fix memory leak

rbuj:


raveit65 (approved):
LGTM,
builds and runs fine.

PR #600, opened 2021-01-17
multiload: fix netspeed_add signature

hvdijk:
Commit c85a5cf5 changed the definition to use guint64 rather than
gulong, but did not update the declaration in netspeed.h to match.


rbuj (approved):
LGTM

PR #609, opened 2021-09-02
Workaround to fix #372

### LINKED ISSUE #372: Charpick (Character Palette) applet overlapping
Hello! I have a small bug with the Character Palette applet.

#### Expected behaviour
Normal display of buttons when switching them

screenshottraditional
(TraditionalOk desktop theme)

#### Actual behaviour
When switching sets, the buttons overlap each other.

screenshot2
screenshot3
screenshot

1. Add Character Palette applet to panel
2. Choose one character set
3. Click on any char
4. Choose other character set

This bug appear in all desktop themes (from mate-themes) except TraditionalOk and TraditionalGreen.

#### MATE general version
1.20.3

#### Package version
mate-applets v1.20.2

#### Linux Distribution
ALT Linux Workstation 8.2
Mint 19

balping:
This forces the background of the applet to be re-rendered thus preventing overlapping images.

### DIFF
--- a/charpick/charpick.c
+++ b/charpick/charpick.c
@@ -403,6 +403,22 @@ build_table (charpick_data *p_curr_data)
     else
         box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 0);
 
+    /* Workaround to fix overlapping bug #372 */
+    /* FIXME: actual bug is probably at a deeper level in GTK */
+    GtkCssProvider *provider;
+    provider = gtk_css_provider_new ();
+    gtk_css_provider_load_from_data (provider,
+                                     "#charpick-applet {\n"
+                                     "background-color:@theme_bg_color;\n"
+                                     "}",
+                                     -1,
+                                     NULL);
+    gtk_style_context_add_provider (gtk_widget_get_style_context (box),
+                                    GTK_STYLE_PROVIDER (provider),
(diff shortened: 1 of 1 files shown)


muktupavels:
If I remember correctly real problem is that applet widgets are changed from `size_allocate`. You are not supposed to destroy and/or create widgets from that callback.

balping:
@muktupavels  Then where should `build_table` be called instead?

muktupavels:
In `g_idle_add` callback?

balping:
> In `g_idle_add` callback?

I tried that now, it doesn't fix the issue.

muktupavels:
Then maybe I fixed this/similar bug in different commit. Maybe this one:   
https://gitlab.gnome.org/GNOME/gnome-applets/-/commit/36f313d880aef4df2925a1fce585fa820429d4fd

And for idle:   
https://gitlab.gnome.org/GNOME/gnome-applets/-/commit/19c7ccdc26094988c979fa2ee9ace315acfa9f5e

Found bug that might or might not be related - https://gitlab.gnome.org/GNOME/gtk/-/issues/1540. Don't remember... And my commit messages are not helpful...

lukefromdc:
Noe that this (closed) fix creates issues in the Yaru themes and BlackMATE. Not a worry as we are moving on

PR #621, opened 2022-01-03
geyes: consider the offset widget <-> window

kreijack:
Problem: when the mouse pointer is between two eyes, the left one is correct oriented; however the right one is wrongly oriented 

image

I discovered that this is due to the fact that it is not computed the offset between the single eye widget and their parent window.

Correct views:
image
image
image
image

To compute the correct pupil position, we should consider the offset
between the eye widget and their parent window.

Signed-off-by: Goffredo Baroncelli <kreijack@inwind.it>

### DIFF
--- a/geyes/src/geyes.c
+++ b/geyes/src/geyes.c
@@ -132,7 +132,7 @@ timer_cb (EyesApplet *eyes_applet)
 {
     GdkDisplay *display;
     GdkSeat *seat;
-    gint x, y;
+    gint x, y, dx, dy;
     gint pupil_x, pupil_y;
     gsize i;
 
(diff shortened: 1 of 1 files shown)


rbuj (approved):
I only see a formatting issue: missing space between the function name and the round bracket. Other than that, it seems right to me as the [convergence](https://en.wikipedia.org/wiki/Vergence) is a valid eye movement.

kreijack:
Great !!
Do you want that I send an updated patch to solve the formatting issue ?

rbuj:
You don't need to make a new merge request. You can modify the file and upload the changes to the current commit.

raveit65 (approved):
LGTM

raveit65:
Thank you

raveit65:
I will consider this as a bug fix and push it to 1.26 branch.

PR #625, opened 2022-02-07
multiload: fix memory leak

rbuj:


mbkma (approved):
LGTM, `g_settings_get_string` doesn't return NULL so its safe to dereference. And `color_string` gets allocated in any case, so it must be freed not only in the if statement.

lukefromdc (approved):
The multiload (system monitor) applet works fine with this installed, and graph colors can be changed as before

PR #631, opened 2022-07-24
command: fix -Wincompatible-pointer-types warning

rbuj:
```
command.c:472:45: warning: incompatible pointer types passing 'MatePanelApplet *' (aka 'struct _MatePanelApplet *') to parameter of type 'GtkWidget *' (aka 'struct _GtkWidget *') [-Wincompatible-pointer-types]
    atk_widget = gtk_widget_get_accessible (command_applet->applet);
                                            ^~~~~~~~~~~~~~~~~~~~~~
/usr/include/gtk-3.0/gtk/gtkwidget.h:1067:79: note: passing argument to parameter 'widget' here
AtkObject*       gtk_widget_get_accessible               (GtkWidget          *widget);
                                                                              ^
```

### DIFF
--- a/command/src/command.c
+++ b/command/src/command.c
@@ -469,7 +469,7 @@ command_applet_fill (MatePanelApplet* applet)
                      "visible",
                      G_SETTINGS_BIND_DEFAULT);
 
-    atk_widget = gtk_widget_get_accessible (command_applet->applet);
+    atk_widget = gtk_widget_get_accessible (GTK_WIDGET (command_applet->applet));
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
With this applied we get only deprecation warnings from this file.Builds and runs fine with indistingushable behavior in normal (non-ORCA) use.  While I have NOT tested with Orca, accerciser picks up the "command applet" name and the contents of the output label. Thus, this should also work fine with Orca. No reason it should not anyway as this just applies a macro to getting the atk widget

zhuyaliang (approved):

PR #632, opened 2022-07-25
configure.ac: The macro 'AC_HELP_STRING' is obsolete

rbuj:


lukefromdc:
AS_HELP_STRING has been around back to 2008 at LEAST, as that's when Autoconf 2.62 came out. Thus no distro should get a build failure from this change

lukefromdc:
About to test this on Debian myself, the cancelled test is also for Debian. Getting browser/3ed party content issues from my filtering software on trying to rerun that, so I will simply substituter my own test build.

lukefromdc (approved):
`./configure --help` after 
`./autogen.sh` works fine with this on Debian Unstable, as does a build. Will merge as we now know this works on Debian

raveit65:
@lukefromdc 
Debian travis-Ci build is broken here (2 of 3 checks passed) and in master.
Not shure if this commit is the culbrit or this one https://github.com/mate-desktop/mate-applets/commit/8c06942386ee04905eeee309bea9a5e0284e7c9e

lukefromdc:
Right now, the only way I can stop these travis problems is to entirely stop merging anything. I don't know enough about travis, working in containers, or .yml to understand this.  

Wondering though if some change has simply broken ALL new Debian builds regardless of code changes, as the Fedora builds seem to finish OK.

lukefromdc:
Sure test would be to run travis on a branch with these reverted and see what happens: if the builds still break, the problem is elsewhere. How would I do that?

lukefromdc:
I went through the 
.build.yml file and did find the specification for gtksourceview 4.0 in it
 debian:
```
    # Useful URL: https://github.com/mate-desktop/debian-packages
    # Useful URL: https://salsa.debian.org/debian-mate-team/mate-applets
    - autopoint
    - clang
    - clang-tools
    - cppcheck
    - gcc
    - git
    - libcpupower-dev
    - libdbus-1-dev
    - libdbus-glib-1-dev
    - libglib2.0-dev
    - libgtk-3-dev
    - libgtksourceview-4-dev
    - ...
```
So the build failure on travis isn't cause by specifying the wrong dependency there

raveit65:
For me this looks looks like a bug in debian development (debian-testing) or they did change something which causes a build error.
All what we can do is to try to understand this build issues, eg. reproducing with a local build. In PR the build was OK. So it might be an issue with current docker image from debian. They will be updated after a while. If debian build is broken after some weeks than it is definitely our problem.
The tag Fedora:latest (f36) is a stable release, which indicates that something is broken in debian-testing or new.

lukefromdc:
We don't have this AC_HELP_STRING>AS_HELP_STRING change in caja-extensions, that build succeeded. On the other hand, we also don't have any warning messages related to either one, and neither should have any effect unless `./configure --help` is run

raveit65:
`Compilation failed: Failed to close file descriptor for child process (Operation not permitted)`
Seems that is a issue with the current docker image.

PR #635, opened 2022-08-06
stickynotes: bump gtksourceview 4

### LINKED ISSUE #634: mate-applets still uses gtksourceview:3.0 while pluma uses gtksourceview:4.0
#### Expected behaviour
Wold be nice if MATE used the same  gtksourceview throughout

#### Actual behaviour
Use gtksourceview:3.0

#### MATE general version
1.26

#### Package version
1.26.0

#### Linux Distribution
Gentoo

#### Link to bugreport of your Distribution (requirement)
Not a dist bug

rbuj:
GtkSourceBuffer provides the ability to undo or redo the change to the note text-body.  [pluma](https://github.com/mate-desktop/pluma/commit/69870af1aed34b32a48ff75554b4438e222cc526) uses gtksourceview 4, so we don't need to depend on two versions.

fix #634

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -10,7 +10,7 @@ requires:
     - gcc
     - git
     - gucharmap
-    - gtksourceview3
+    - gtksourceview4
     - itstool
     - libgtop
     - libnotify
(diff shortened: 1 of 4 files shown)


lukefromdc (approved):
Builds and runs fine. Undoing and redoing test in a sticky note works without issue with this installed.  This has the advantage of removing from MATE as a whole a dependency on two versions of the same library unless a 3ed MATE application is using it, in which case it is a step in that direction.

lukefromdc:
Merging this: it works, came from a trusted contributor, doesn't seem to be attracting any other reviewers, and reduces our dependency footprint for distros

PR #638, opened 2022-08-07
cpufreq: Add data dir

rbuj:


lukefromdc (approved):
cpufreq-applet builds and runs fine with this change applied. Didn't notice any build warnings about the new directory, just the usual deprecation warnings

PR #639, opened 2022-08-07
trashapplet: Add data dir

rbuj:


lukefromdc (approved):
Trash applet builds and runs fine with this applied. I didn't notice any build warnings related to the new data directory either

PR #648, opened 2023-02-08
trashapplet: Replace the deprecated function

zhuyaliang:
Use ```g_task``` replaces the deprecated ```g_io_scheduler```, removing compilation warnings
```
trash-empty.c: In function 'trash_empty_job':
trash-empty.c:217:3: warning: 'g_io_scheduler_job_send_to_mainloop_async' is deprecated: Use 'g_main_context_invoke' instead [-Wdeprecated-declarations]
   g_io_scheduler_job_send_to_mainloop_async (job,
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/glib-2.0/gio/gio.h:89,
                 from trash-empty.c:22:
/usr/include/glib-2.0/gio/gioscheduler.h:47:10: note: declared here
 void     g_io_scheduler_job_send_to_mainloop_async (GIOSchedulerJob     *job,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
trash-empty.c: In function 'trash_empty_start':
trash-empty.c:267:3: warning: 'g_io_scheduler_push_job' is deprecated: Use '"GThreadPool or g_task_run_in_thread"' instead [-Wdeprecated-declarations]
   g_io_scheduler_push_job (trash_empty_job, NULL, NULL, 0, cancellable);
```

### DIFF
--- a/trashapplet/src/trash-empty.c
+++ b/trashapplet/src/trash-empty.c
@@ -114,20 +114,19 @@ trash_empty_update_dialog (gpointer user_data)
   return FALSE;
 }
 
-static gboolean
-trash_empty_done (gpointer user_data)
+static void
+trash_empty_done (GObject      *source_object,
+                  GAsyncResult *res,
+                  gpointer      user_data)
 {
   gtk_widget_destroy (GTK_WIDGET (trash_empty_dialog));
 
   g_assert (trash_empty_dialog == NULL);
-
-  return FALSE;
 }
 
 /* =============== worker thread code begins here =============== */
 static void
-trash_empty_maybe_schedule_update (GIOSchedulerJob *job,
-                                   GFile           *file,
+trash_empty_maybe_schedule_update (GFile           *file,
                                    gsize            deleted)
 {
   if (!trash_empty_update_pending)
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
This removes all deprecation warnings from building the trash applet with default options. Moving items to the trash, emptying the trash and restoring from the trash work same as before. Note that there is a memory leak somewhere in the trash applet but this does not seem to affect it.

I could only find one new heap variable (a GAsyncResult) that is not explicitly unref'ed, and looking at the docs I can't figure out if we are supposed to unref it, so I tested changing trash_empty_done() to:

```
static void
trash_empty_done (GObject      *source_object,
                  GAsyncResult *res,
                  gpointer      user_data)
{
  gtk_widget_destroy (GTK_WIDGET (trash_empty_dialog));
  g_object_unref (res);

  g_assert (trash_empty_dialog == NULL);
}
```
The  ` g_object_unref (res);` is my addition, didn't get any crashes from it but I do NOT know enough about gio to know if we need this or not. It seemed to stop ram usage from rising on emptying the trash from the applet  (which it does both with and without this PR), and didn't cause a crash. The main memory leak though is elsewhere when items are added to the trash, with that ram use not being freed when the trash is emptied nor when the task of adding items to the trash is done.

zhuyaliang:
@lukefromdc The unref been completed here
```
  task = g_task_new (parent, cancellable, trash_empty_done, NULL);
  g_task_run_in_thread (task, trash_empty_job);
  g_object_unref (task);
```

PR #651, opened 2023-06-20
Port all wayland-usable applets to in-process

lukefromdc:
Most mate-panel applets work in wayland the same or nearly the same as on x11 after porting to in-process. 

*Accessx not ported, hard dependency on x11 for keyboard handling. The standalone xkb library could offer a future way to fix this

*The battstat applets works same in x11 and wayland, save that again notifications cannot yet be used

*The charpick applet itself will function in both x11 and wayland, but the primary paste preference does not work in wayland-and even resets the gsettings preference to disable it. Fixing this is probably the job of a wayland session manager. No change in x11 behavior

*Cpufreq and drivemount work same in x11 and wayland as they did in x11 before

*Geyes not ported as the applet cannot see the pointer in wayland once it moves outside the panel due to wayland restrictions on one app accessing another's windows. This is considered a security feature in wayland as it makes clickjacking much harder. Also when in-process, geyes improperly computes pointer position as it becomes relative to the mate-panel window not the GtkPlug window used for out of process.

*mateweather, multiload,netspeed, stickynotes all seem to work the same in wayland as in x11 and no behavioral differences were noticed in (ONE user) testing

*The timer applet must be used with the dialog not the notification in wayland, as we don't have a wayland-compatable notification daemon yet and the applet will freeze on a failed notification. No change x11 behavior.

*Trash applet works same in x11 and wayland if a wayland-compatable file manager is installed

### DIFF
--- a/accessx-status/src/applet.c
+++ b/accessx-status/src/applet.c
@@ -1316,7 +1316,9 @@ create_applet (MatePanelApplet* applet)
     GtkIconTheme *icon_theme;
     gint icon_size, icon_scale;
 
+#ifndef ENABLE_IN_PROCESS
     g_set_application_name (_("AccessX Status"));
+#endif
 
     sapplet->xkb = NULL;
     sapplet->xkb_display = NULL;
@@ -1633,9 +1635,9 @@ accessx_status_applet_fill (MatePanelApplet* applet)
     }
 
     g_object_connect (sapplet->applet,
-                      "destroy", accessx_status_applet_destroy, sapplet,
-                      "change-orient", accessx_status_applet_reorient, sapplet,
-                      "change-size", accessx_status_applet_resize, sapplet,
+                      "signal::destroy", accessx_status_applet_destroy, sapplet,
+                      "signal::change-orient", accessx_status_applet_reorient, sapplet,
+                      "signal::change-size", accessx_status_applet_resize, sapplet,
                       NULL);
 
     g_signal_connect (sapplet->applet, "button-press-event",
@@ -1697,9 +1699,9 @@ accessx_status_applet_factory (MatePanelApplet* applet,
     return retval;
 }
 
-MATE_PANEL_APPLET_OUT_PROCESS_FACTORY ("AccessxStatusAppletFactory",
-                                       PANEL_TYPE_APPLET,
-                                       "accessx-status",
-                                       accessx_status_applet_factory,
-                                       NULL)
+PANEL_APPLET_FACTORY ("AccessxStatusAppletFactory",
+                      PANEL_TYPE_APPLET,
+                      "accessx-status",
+                      accessx_status_applet_factory,
+                      NULL)
 
--- a/geyes/src/geyes.c
+++ b/geyes/src/geyes.c
@@ -140,8 +140,10 @@ timer_cb (EyesApplet *eyes_applet)
     display = gtk_widget_get_display (GTK_WIDGET (eyes_applet->applet));
     seat = gdk_display_get_default_seat (display);
 
-    for (i = 0; i < eyes_applet->num_eyes; i++) {
-        if (gtk_widget_get_realized (eyes_applet->eyes[i])) {
+    for (i = 0; i < eyes_applet->num_eyes; i++)
+    {
+        if (gtk_widget_get_realized (eyes_applet->eyes[i]))
+        {
             gdk_window_get_device_position (gtk_widget_get_window (eyes_applet->eyes[i]),
                                             gdk_seat_get_pointer (seat),
                                             &x, &y, NULL);
--- a/geyes/src/geyes.c
+++ b/geyes/src/geyes.c
@@ -135,6 +135,7 @@ timer_cb (EyesApplet *eyes_applet)
     gint x, y, dx, dy;
     gint pupil_x, pupil_y;
     gsize i;
+    int applet_x,applet_y;
 
     display = gtk_widget_get_display (GTK_WIDGET (eyes_applet->applet));
     seat = gdk_display_get_default_seat (display);
@@ -144,9 +145,17 @@ timer_cb (EyesApplet *eyes_applet)
             gdk_window_get_device_position (gtk_widget_get_window (eyes_applet->eyes[i]),
                                             gdk_seat_get_pointer (seat),
                                             &x, &y, NULL);
+
+#ifdef ENABLE_IN_PROCESS
+            gdk_window_get_origin (gtk_widget_get_window (GTK_WIDGET(eyes_applet->applet)),
+                                       &applet_x, &applet_y);
+#else
+            applet_x = 0;
+            applet_y = 0;
+#endif
             gtk_widget_translate_coordinates (eyes_applet->eyes[i],
                                               gtk_widget_get_toplevel(eyes_applet->eyes[i]),
-                                              0, 0, &dx, &dy);
+                                              -applet_x, -applet_y, &dx, &dy);
             x -= dx;
             y -= dy;
 
@@ -257,8 +266,6 @@ setup_eyes (EyesApplet *eyes_applet)
             gtk_widget_set_halign (eyes_applet->eyes[i], GTK_ALIGN_CENTER);
             gtk_widget_set_valign (eyes_applet->eyes[i], GTK_ALIGN_CENTER);
         }
-
-        gtk_widget_realize (eyes_applet->eyes[i]);
 
         eyes_applet->pointer_last_x[i] = G_MAXINT;
         eyes_applet->pointer_last_y[i] = G_MAXINT;
@@ -398,7 +405,9 @@ geyes_applet_fill (MatePanelApplet *applet)
     GtkActionGroup *action_group;
     gboolean result;
 
+#ifndef ENABLE_IN_PROCESS
     g_set_application_name (_("Eyes"));
+#endif
     gtk_window_set_default_icon_name ("mate-eyes-applet");
     mate_panel_applet_set_flags (applet, MATE_PANEL_APPLET_EXPAND_MINOR);
 
@@ -466,8 +475,8 @@ geyes_applet_factory (MatePanelApplet *applet,
     return retval;
 }
 
-MATE_PANEL_APPLET_OUT_PROCESS_FACTORY ("GeyesAppletFactory",
-                                       PANEL_TYPE_APPLET,
-                                       "geyes",
-                                       geyes_applet_factory,
-                                       NULL)
+PANEL_APPLET_FACTORY ("GeyesAppletFactory",
+                      PANEL_TYPE_APPLET,
+                      "geyes",
+                      geyes_applet_factory,
+                      NULL)
--- a/netspeed/src/Makefile.am
+++ b/netspeed/src/Makefile.am
@@ -1,8 +1,5 @@
 NULL =
 
-mate_netspeed_applet_libdir= $(pkglibdir)
-mate_netspeed_applet_lib_LTLIBRARIES=libmate-netspeed-applet.la
-
 AM_CPPFLAGS =			\
 	-I$(top_srcdir)		\
 	-I$(includedir)		\
@@ -16,34 +13,53 @@ if HAVE_NL
 AM_CPPFLAGS += $(NL_CFLAGS)
 endif
 
-libmate_netspeed_applet_la_SOURCES =	\
+BUILT_SOURCES =			\
+	netspeed-resources.c	\
+	netspeed-resources.h	\
+	$(NULL)
+APPLET_SOURCES =		\
 	backend.h		\
 	backend.c		\
 	netspeed.c		\
 	netspeed.h		\
 	netspeed-preferences.c	\
 	netspeed-preferences.h	\
-	netspeed-resources.c   \
-	netspeed-resources.h   \
 	$(NULL)
 
 if HAVE_NL
-libmate_netspeed_applet_la_SOURCES +=	\
+APPLET_SOURCES +=		\
 	nl80211.h		\
 	ieee80211.h		\
 	$(NULL)
 endif
 
-libmate_netspeed_applet_la_LIBADD = $(GIO_LIBS) $(GTOP_APPLETS_LIBS) \
-       $(MATE_APPLETS4_LIBS) \
-       $(INTLLIBS) -lm
+APPLET_LIBS = \
+	$(GIO_LIBS) \
+	$(GTOP_APPLETS_LIBS) \
+	$(MATE_APPLETS4_LIBS) \
+	$(INTLLIBS) -lm
 
 if HAVE_IW
-libmate_netspeed_applet_la_LIBADD += $(IW_LIBS)
+APPLET_LIBS += $(IW_LIBS)
 endif
 if HAVE_NL
-libmate_netspeed_applet_la_LIBADD += $(NL_LIBS)
+APPLET_LIBS += $(NL_LIBS)
 endif
+
+if ENABLE_IN_PROCESS
+pkglib_LTLIBRARIES = libmate-netspeed-applet.la
+nodist_libmate_netspeed_applet_la_SOURCES = $(BUILT_SOURCES)
+libmate_netspeed_applet_la_SOURCES = $(APPLET_SOURCES)
+libmate_netspeed_applet_la_CFLAGS = $(AM_CFLAGS)
+libmate_netspeed_applet_la_LDFLAGS = -module -avoid-version
+libmate_netspeed_applet_la_LIBADD = $(APPLET_LIBS)
+else !ENABLE_IN_PROCESS
+libexec_PROGRAMS = mate-netspeed-applet
+nodist_mate_netspeed_applet_SOURCES = $(BUILT_SOURCES)
+mate_netspeed_applet_SOURCES = $(APPLET_SOURCES)
+mate_netspeed_applet_CFLAGS = $(AM_CFLAGS)
+mate_netspeed_applet_LDADD = $(APPLET_LIBS)
+endif !ENABLE_IN_PROCESS
 
 netspeed-resources.c: $(srcdir)/../data/netspeed-resources.gresource.xml $(shell $(GLIB_COMPILE_RESOURCES) --sourcedir=$(srcdir)/../data --generate-dependencies $(srcdir)/../data/netspeed-resources.gresource.xml)
 	$(AM_V_GEN)$(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir)/../data --generate --c-name netspeed $<
@@ -52,6 +68,7 @@ netspeed-resources.h: $(srcdir)/../data/netspeed-resources.gresource.xml $(shell
 	$(AM_V_GEN)$(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir)/../data --generate --c-name netspeed $<
 
 CLEANFILES =			\
+	$(BUILT_SOURCES)	\
 	$(NULL)
 
 -include $(top_srcdir)/git.mk
--- a/netspeed/src/Makefile.am
+++ b/netspeed/src/Makefile.am
@@ -1,5 +1,8 @@
 NULL =
 
+mate_netspeed_applet_libdir= $(pkglibdir)
+mate_netspeed_applet_lib_LTLIBRARIES=libmate-netspeed-applet.la
+
 AM_CPPFLAGS =			\
 	-I$(top_srcdir)		\
 	-I$(includedir)		\
@@ -13,35 +16,33 @@ if HAVE_NL
 AM_CPPFLAGS += $(NL_CFLAGS)
 endif
 
-libexec_PROGRAMS = mate-netspeed-applet
-
-BUILT_SOURCES = netspeed-resources.c netspeed-resources.h
-nodist_mate_netspeed_applet_SOURCES = $(BUILT_SOURCES)
-mate_netspeed_applet_SOURCES =	\
+libmate_netspeed_applet_la_SOURCES =	\
 	backend.h		\
 	backend.c		\
 	netspeed.c		\
 	netspeed.h		\
 	netspeed-preferences.c	\
 	netspeed-preferences.h	\
+	netspeed-resources.c   \
+	netspeed-resources.h   \
 	$(NULL)
 
 if HAVE_NL
-mate_netspeed_applet_SOURCES +=	\
+libmate_netspeed_applet_la_SOURCES +=	\
 	nl80211.h		\
 	ieee80211.h		\
 	$(NULL)
 endif
 
-mate_netspeed_applet_LDADD = $(GIO_LIBS) $(GTOP_APPLETS_LIBS) \
+libmate_netspeed_applet_la_LIBADD = $(GIO_LIBS) $(GTOP_APPLETS_LIBS) \
        $(MATE_APPLETS4_LIBS) \
        $(INTLLIBS) -lm
 
 if HAVE_IW
-mate_netspeed_applet_LDADD += $(IW_LIBS)
+libmate_netspeed_applet_la_LIBADD += $(IW_LIBS)
 endif
 if HAVE_NL
-mate_netspeed_applet_LDADD += $(NL_LIBS)
+libmate_netspeed_applet_la_LIBADD += $(NL_LIBS)
 endif
 
 netspeed-resources.c: $(srcdir)/../data/netspeed-resources.gresource.xml $(shell $(GLIB_COMPILE_RESOURCES) --sourcedir=$(srcdir)/../data --generate-dependencies $(srcdir)/../data/netspeed-resources.gresource.xml)
@@ -51,7 +52,6 @@ netspeed-resources.h: $(srcdir)/../data/netspeed-resources.gresource.xml $(shell
 	$(AM_V_GEN)$(GLIB_COMPILE_RESOURCES) --target=$@ --sourcedir=$(srcdir)/../data --generate --c-name netspeed $<
 
 CLEANFILES =			\
-	$(BUILT_SOURCES)	\
 	$(NULL)
 
 -include $(top_srcdir)/git.mk
--- a/stickynotes/stickynotes.c
+++ b/stickynotes/stickynotes.c
@@ -809,12 +809,10 @@ stickynotes_remove (StickyNote *note)
 gboolean
 stickynotes_save_now (void)
 {
-    WnckScreen *wnck_screen;
     const gchar *title;
     GtkTextBuffer *buffer;
     GtkTextIter start, end;
     gchar *body;
-
     gint i;
 
     /* Create a new XML document */
@@ -823,14 +821,16 @@ stickynotes_save_now (void)
 
     xmlDocSetRootElement (doc, root);
     xmlNewProp (root, XML_CHAR ("version"), XML_CHAR (VERSION));
-
-    wnck_screen = wnck_screen_get_default ();
-    wnck_screen_force_update (wnck_screen);
-
+#ifdef GDK_WINDOWING_X11
+    GdkDisplay *display = gdk_screen_get_display(gdk_screen_get_default());
+    if (GDK_IS_X11_DISPLAY(display))
+    {
+        WnckScreen *wnck_screen = wnck_screen_get_default ();
+        wnck_screen_force_update (wnck_screen);
+    }
+#endif
     /* For all sticky notes */
     for (i = 0; i < g_list_length (stickynotes->notes); i++) {
-        WnckWindow *wnck_win;
-        gulong xid = 0;
 
         /* Access the current note in the list */
         StickyNote *note = g_list_nth_data (stickynotes->notes, i);
@@ -842,16 +842,19 @@ stickynotes_save_now (void)
         /* Retrieve the window position of the note */
         gchar *x_str = g_strdup_printf ("%d", note->x);
         gchar *y_str = g_strdup_printf ("%d", note->y);
+#ifdef GDK_WINDOWING_X11
+        if (GDK_IS_X11_DISPLAY(display))
+        {
+            gulong xid = GDK_WINDOW_XID (gtk_widget_get_window (note->w_window));
+            WnckWindow *wnck_win = wnck_window_get (xid);
 
-        xid = GDK_WINDOW_XID (gtk_widget_get_window (note->w_window));
-        wnck_win = wnck_window_get (xid);
-
-        if (!g_settings_get_boolean (stickynotes->settings, "sticky") && wnck_win)
-            note->workspace = 1 +
-                wnck_workspace_get_number (wnck_window_get_workspace (wnck_win));
-        else
-            note->workspace = 0;
-
+            if (!g_settings_get_boolean (stickynotes->settings, "sticky") && wnck_win)
+                note->workspace = 1 +
+                    wnck_workspace_get_number (wnck_window_get_workspace (wnck_win));
+            else
+                note->workspace = 0;
+        }
+#endif
         /* Retrieve the title of the note */
         title = gtk_label_get_text (GTK_LABEL (note->w_title));
 
@@ -948,10 +951,12 @@ stickynotes_load (GdkScreen *screen)
     xmlDocPtr doc = NULL;
     xmlNodePtr root;
     xmlNodePtr node;
-    /* WnckScreen *wnck_screen; */
     GList *new_notes, *tmp1;  /* Lists of StickyNote*'s */
     GList *new_nodes; /* Lists of xmlNodePtr's */
     int x, y, w, h;
+#ifdef GDK_WINDOWING_X11
+    GdkDisplay *display = gdk_screen_get_display(gdk_screen_get_default());
+#endif
 
     /* The XML file is $HOME/.config/mate/stickynotes-applet, most probably */
     gchar* file = g_build_filename (g_get_user_config_dir (),
@@ -1123,11 +1128,14 @@ stickynotes_load (GdkScreen *screen)
     }
 
     tmp1 = new_notes;
-    /*
-    wnck_screen = wnck_screen_get_default ();
-    wnck_screen_force_update (wnck_screen);
-    */
 
+#ifdef GDK_WINDOWING_X11
+    if (GDK_IS_X11_DISPLAY(display))
+    {
+        WnckScreen *wnck_screen = wnck_screen_get_default ();
+        wnck_screen_force_update (wnck_screen);
+    }
+#endif
     while (tmp1) {
         StickyNote *note = tmp1->data;
 
--- a/stickynotes/stickynotes.c
+++ b/stickynotes/stickynotes.c
@@ -822,8 +822,8 @@ stickynotes_save_now (void)
     xmlDocSetRootElement (doc, root);
     xmlNewProp (root, XML_CHAR ("version"), XML_CHAR (VERSION));
 #ifdef GDK_WINDOWING_X11
-    GdkDisplay *display = gdk_screen_get_display(gdk_screen_get_default());
-    if (GDK_IS_X11_DISPLAY(display))
+    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
+    if (GDK_IS_X11_DISPLAY (display))
     {
         WnckScreen *wnck_screen = wnck_screen_get_default ();
         wnck_screen_force_update (wnck_screen);
@@ -955,7 +955,7 @@ stickynotes_load (GdkScreen *screen)
     GList *new_nodes; /* Lists of xmlNodePtr's */
     int x, y, w, h;
 #ifdef GDK_WINDOWING_X11
-    GdkDisplay *display = gdk_screen_get_display(gdk_screen_get_default());
+    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
 #endif
 
     /* The XML file is $HOME/.config/mate/stickynotes-applet, most probably */
--- a/trashapplet/data/Makefile.am
+++ b/trashapplet/data/Makefile.am
@@ -1,32 +1,25 @@
+APPLET_LOCATION = $(libdir)/mate-applets/libtrashapplet.so
+
 appletdir       = $(datadir)/mate-panel/applets
 applet_in_files = org.mate.applets.TrashApplet.mate-panel-applet.desktop.in
 applet_DATA     = $(applet_in_files:.mate-panel-applet.desktop.in=.mate-panel-applet)
 
 $(applet_in_files): $(applet_in_files).in Makefile
 	$(AM_V_GEN)sed \
-            -e "s|\@LIBEXECDIR\@|$(libexecdir)|" \
-            -e "s|\@VERSION\@|$(PACKAGE_VERSION)|" \
-            $< > $@
+		-e "s|\@APPLET_LOCATION\@|$(APPLET_LOCATION)|" \
+		-e "s|\@VERSION\@|$(PACKAGE_VERSION)|" \
+		$< > $@
 
 $(applet_DATA): $(applet_in_files) Makefile
 	$(AM_V_GEN) $(MSGFMT) --desktop --keyword=Name --keyword=Description --template $< -d $(top_srcdir)/po -o $@
 
-servicedir       = $(datadir)/dbus-1/services
-service_in_files = org.mate.panel.applet.TrashAppletFactory.service.in
-service_DATA     = $(service_in_files:.service.in=.service)
-
-org.mate.panel.applet.TrashAppletFactory.service: $(service_in_files)
-	$(AM_V_GEN)sed \
-            -e "s|\@LIBEXECDIR\@|$(libexecdir)|" \
-            $< > $@
 
 EXTRA_DIST = \
 	$(applet_in_files).in	\
-	$(service_in_files) \
 	trashapplet-empty-progress.ui \
 	trashapplet-menu.xml \
 	trashapplet-resources.gresource.xml
 
-CLEANFILES = $(applet_DATA) $(applet_in_files) $(service_DATA)
+CLEANFILES = $(applet_DATA) $(applet_in_files)
 
 -include $(top_srcdir)/git.mk
--- a/trashapplet/data/Makefile.am
+++ b/trashapplet/data/Makefile.am
@@ -1,25 +1,42 @@
-APPLET_LOCATION = $(libdir)/mate-applets/libtrashapplet.so
+applet_in_files = org.mate.applets.TrashApplet.mate-panel-applet.desktop.in
+service_in_files = org.mate.panel.applet.TrashAppletFactory.service.in
+
+if ENABLE_IN_PROCESS
+APPLET_LOCATION = $(pkglibdir)/libmate-trash-applet.so
+else !ENABLE_IN_PROCESS
+APPLET_LOCATION = $(libexecdir)/trashapplet
+endif !ENABLE_IN_PROCESS
 
 appletdir       = $(datadir)/mate-panel/applets
-applet_in_files = org.mate.applets.TrashApplet.mate-panel-applet.desktop.in
 applet_DATA     = $(applet_in_files:.mate-panel-applet.desktop.in=.mate-panel-applet)
 
 $(applet_in_files): $(applet_in_files).in Makefile
 	$(AM_V_GEN)sed \
 		-e "s|\@APPLET_LOCATION\@|$(APPLET_LOCATION)|" \
+		-e "s|\@APPLET_IN_PROCESS\@|$(APPLET_IN_PROCESS)|" \
 		-e "s|\@VERSION\@|$(PACKAGE_VERSION)|" \
 		$< > $@
 
 $(applet_DATA): $(applet_in_files) Makefile
 	$(AM_V_GEN) $(MSGFMT) --desktop --keyword=Name --keyword=Description --template $< -d $(top_srcdir)/po -o $@
 
+if !ENABLE_IN_PROCESS
+servicedir = $(datadir)/dbus-1/services
+service_DATA = $(service_in_files:.service.in=.service)
 
-EXTRA_DIST = \
-	$(applet_in_files).in	\
-	trashapplet-empty-progress.ui \
-	trashapplet-menu.xml \
+$(service_DATA): $(service_in_files) Makefile
+	$(AM_V_GEN)sed \
+		-e "s|\@APPLET_LOCATION\@|$(APPLET_LOCATION)|" \
+		$< > $@
+endif !ENABLE_IN_PROCESS
+
+EXTRA_DIST =					\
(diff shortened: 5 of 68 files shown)


lukefromdc:
Looks like an issue running a docker script not the build itself failing on the Fedora build by travis

raveit65:
No, this is an library linking error.
https://app.travis-ci.com/github/mate-desktop/mate-applets/jobs/604569199#L4989
```
*** Warning: Linking the shared library libmate-battstat-applet.la against the

*** static library apmlib/libapm.a is not portable!

/usr/bin/ld: apmlib/libapm.a(apmlib.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC

/usr/bin/ld: failed to set dynamic section sizes: bad value

collect2: error: ld returned 1 exit status

make[3]: *** [Makefile:617: libmate-battstat-applet.la] Error 1

make[3]: *** Waiting for unfinished jobs....

make[3]: Leaving directory '/rootdir/battstat'

make[2]: *** [Makefile:694: all-recursive] Error 1

make[2]: Leaving directory '/rootdir/battstat'

make[1]: *** [Makefile:583: all-recursive] Error 1

make[1]: Leaving directory '/rootdir'

make: *** [Makefile:515: all] Error 2
```
`ld` failed to run.

lukefromdc:
This is a primitive mate-wayland session running, w caja managing the desktop and some of these panel
applets in use in my normal configuration. Photo taken w camera due to lack of a screenshot utility for wayland
on my system at this time

Mate-Wayland_6--20-2023

cwendling:
@raveit65 @lukefromdc check out #652, it should help with the libapm PIC issue.  The reason why the problem appears now is probably because the applet is now in-process, and thus a library (requiring PIC on many platforms, including x86_64), instead of a separate executable (which doesn't usually need PIC).  Using libtool should fix all cases, as it is able to build both static and dynamic libraries, suited to the platform.

raveit65:
Just rebased with master including the fix for battstat-applet.

lukefromdc:
That came from not knowing enough about the build system-and from not realizinng that library was part of the package and thus fixable without having to have another package change. Rebuilding now

lukefromdc:
Just cleaned up that last commit message since a fix is in master for the warning

lukefromdc:
Battstat now builds with only deprecation warnings, thanks!

trashapplet/src/trash-empty.c
```diff
@@ -229,8 +229,10 @@ trash_empty_start (GtkWidget *parent)
   GtkBuilder *builder;
   GTask      *task;
   gint i;
+  const gchar *resource_name;
 
-  builder = gtk_builder_new_from_resource (GRESOURCE "trashapplet-empty-progress.ui");
+  resource_name = "/org/gnome/gnome-applets/trash/trashapplet-empty-progress.ui";
```
raveit65, line 234: gnome-applets? really?

cwendling: Yeah obviously should be `/org/mate/mate-applets/trash/trashapplet-empty-progress.ui`.

@lukefromdc any reason why you chose to drop the `GRESOURCE` macro?  It's not so pretty, but it allows to unify the prefix used in both places.  I don't care much, but it would likely not have lead to this :)

lukefromdc: That looks like an accident, trash was the first applet I ported. I will bring that back and see if I can clean up the makefile.am files  a bit too

lukefromdc: Starting point was gnome-applet's 2016 port to in-process, so "gnome-applets" is a bad cut and paste. Thanks for catching this!

lukefromdc: Brought it back.

raveit65:
Starting with trash-applet.
Everything seems OK in x11.
In wayfire:
- it displays well at the panel 
- icon seems to be scaled with HIDPI (not 100% shure)
- no icons in menu
- open-trash menu-entry doesn't work

raveit65:
Command-applet:
- works fine under x11
wayfire:
- it displays at the panel but icon isn't scaled with HIDPI. It looks blurred.
- function is there
- no icons in menu
- function is there

raveit65:
Drivemount applet:
Runs fine under x11 when build in-process.
Wayfire:
- displays fine in the panel, icons are looking same as with x11
- function seems Ok, i could mount a windows partition. For my linux file-sytems  the user permissions are missing for my wayland-test-user
- icons in menus are missing

raveit65:
Weather-applet:
Works fine under x11 when build in-process
Wayfire:
- Displays fine in panel and icon and text seems to be scaled with HIDPI
- no icons in menus
- function is there

raveit65:
Multiload-applet:
Works fine under x11 when build in-process
Wayfire:
- displays fine on the panel, looks Ok.
- function is there
- only applet-menu icons are missing

raveit65:
Netspeed-applet:
Works fine under x11 when build in-process
-Wayfire:
- It displays on the panel, but icon is really blurred with HIDPI, text isn't really sharp.
- function is there
- icons are missing in applet-menus

raveit65:
Sticky-notes:
Works fine under x11 when build in-process
Wayfire:
- displays on the panel but icon is blurred with HIDPI scaling
- function is limited because notes are a x-wayland window 
- icons are missing in applet-menu

raveit65:
Timer-applet:
Works fine under x11 when build in-process
Wayfire:
- displays fine on the panel and icon plus text seems to be Ok with scaling (not really sure)
- function OK
- icons are missing in applet-menu

raveit65:
In general it seems port to in-process is OK. We need to discuss if we need an option to build in-process/out-of-process.

There are 2 general issues for me. I am thinking both are mate-panel bugs after port to wayland. Icons in applet menus are missing and i thing we lost scaling for HIDPI. Which is a huge step back for me compare to my clear and sharp x11 HIDPI desktop. Btw, maybe x11 do better scaling than wayland in general? Here it looks like it.

At least some applets needs some extra work for wayland in single future PRs.

lukefromdc:
The no menu icons issue is a GTK default, remember no session manager is running. This is true also of button icons, which is why I had to force override this in the applet code to make the SNI tray work. For now, adding this code to ~/.config/gtk-3.0/settings.ini will bring the icons back in wayland:
`gtk-menu-images = true`
and adding 
gtk-button-images = true
will make sure those always show up in all GTK apps in wayland

We can either consider that a job for a future wayland settings manager, or modifiy the applets to check gsettings themselves and not wait for a session manager. Remember that mate-panel has the potential to become a popular wayland panel outside of any future MATE session.

Coding all the applets to show the icons unconditionally could be put behind  wayland-only selectors (or !x11 to avoid a new dependency) in every applet here, but would force all users to show the icons. Not sure if the applets can read the gsettings prefs outside any session manager running but I suspect they probably can. This would need to be done for every applet to make the icons follow the user set show/hide preference in a standalone mate-panel in an arbitrary wl-roots supporting wayland session

raveit65:
>> Not sure if the applets can read the gsettings prefs outside any session manager running but I suspect they probably can.

I think so, i am using gnome gsettings keys for gtk-theme, icon-theme and windows-decoration.

lukefromdc:
We need to find out why I can open the trash and you cannot. Are you starting wayfire with 
`dbus-launch wayfire` or something else? On my end caja won't work without that in wayland.

UPDATE: I just had a successful open of trash in caja with caja NOT running. 
I assume you are testing with the wayland-desktop branch of caja installed?

raveit65:
I am starting wayfire from gdm display-manager. But it makes no different when using `dbus-launch wayfire`
But i think i found the problem:
Debugging with:
```
journalctl --user -b0 -f --no-tail
<cut>
.
Jul 03 18:58:09 mother.mother.loc dbus-daemon[123847]: [session uid=1003 pid=123845] Activating service name='org.gnome.Nautilus' requested by ':1.0' (uid=1003 pid=123883 comm="mate-panel")
Jul 03 18:58:09 mother.mother.loc nautilus[124739]: Connecting to org.freedesktop.Tracker3.Miner.Files
Jul 03 18:58:09 mother.mother.loc dbus-daemon[123847]: [session uid=1003 pid=123845] Successfully activated service 'org.gnome.Nautilus'
Jul 03 18:58:09 mother.mother.loc nautilus[124739]: No such backend: *
Jul 03 18:58:09 mother.mother.loc nautilus[124739]: Failed to open display
```
This happens when i click on open-trash-can with the applet.
Seems, that wayfire use some gnome defaults in fedora and i ran in a config issue.
I will dig deeper.

lukefromdc:
I've had problems in the past with the trash opening the wrong file manager even inside a MATE session,
due to some misconfiguration or another

raveit65:
> UPDATE: I just had a successful open of trash in caja with caja NOT running.
> I assume you are testing with the wayland-desktop branch of caja installed?

I use this in wayfire.ini to start caja.
```
[autostart]
a0 = caja --force-desktop -n
a1 = nm-applet --indicator
autostart_wf_shell = false
background = wf-background
gamma = wlsunset
idle = swayidle before-sleep swaylock
notifications = mako
outputs = kanshi
panel = mate-panel
portal = /usr/libexec/xdg-desktop-portal
```
Not sure for what `xdg-desktop-portal` is good for.

raveit65:
Found this link `Boycott Wayland. It breaks everything! ` https://gist.github.com/probonopd/9feb7c20257af5dd915e3a9f2d1f2277

lukefromdc:
That just shows reasons not to go wayland-only as far as I am concerned. We've seen similar calls with systemd and with GTK3 over the years, especially with anything developed by Red Hat or GNOME it seems. 

We do not know what the future holds for wayland vs x11. Being able to use either one is future-proofing and widens the userbase. This does however boost the point that no compromises should be made in the x11 code to support wayland, rather the codepaths should be split.

lukefromdc:
NICE tip about how to start my whole wayland session in wayfire!

raveit65:
Issue with trash-applet i could fix with adding a /home/Sa-wayland/.config/mimeapps.list.
For some reasons this file was missing in home dir.
And i use `portal = /usr/libexec/xdg-desktop-portal-gtk` in wayfire.ini [autostart]
`portal = /usr/libexec/xdg-desktop-portal` seems for flatpak installation.
But the issue with missing icons in applet menus still exists after adding
~/.config/gtk-3.0/settings.ini with that content
```
gtk-menu-images = true
gtk-button-images = true
```
Applets from mate-panel package have icons in menu ie. add-server-applet.
mate-panel_wayfire_aad_to_server_applet
All applets from mate-applets package have missing icons, ie. multiload:
mate-panel_wayfire_multiload_applet
Sorry, i was to lazy to switch to english language in screenshots :)

lukefromdc:
We have a difference in behavior, probably best to just start forcing this with a check of the relevent gsettings prefs applet by applet.

lukefromdc:
Also I get a mix of launcher icons in wayland, some blurry and some not, my guess is my icon theme may be providing the good ones as .svg files

lukefromdc:
Another difference in behavior is cpufreq-applet works here in wayland, but note that I have it set up in x11 not to require passphrase entry to change cpu frequencies. If your is set up to ask for a passphrase, that dialog might contain an x'ism and not be coming up

raveit65:
> Another difference in behavior is cpufreq-applet works here in wayland, but note that I have it set up in x11 not to require passphrase entry to change cpu frequencies. If your is set up to ask for a passphrase, that dialog might contain an x'ism and not be coming up

I use `--enable-polkit` to compile mate-applets in fedora. As distro-maintainer i can't build a package without it.
I guess this flag is for cpu-freq applet. Seems like the polkit daemon isn't running here.

lukefromdc:
Polkit can be set up to allow changing cpu frequency without a passphrase. I've long forgotten how I did that, but it's how I originally did this back in 2015. Never had to fiddle with it. Current build options for mate-applets here are

`./autogen.sh --prefix=/usr libdir=/usr/lib/x86_64-linux-gnu  --sysconfdir=/etc --enable-shared=yes --enable-static=no --disable-maintainer-mode  --disable-schemas-compile --with-cpufreq-lib=cpupower`

The daemon would not be started by my wayland session, so either something else is going on here or in my setup it's being silently started since the passphrase is not called for

raveit65:
> Also I get a mix of launcher icons in wayland, some blurry and some not, my guess is my icon theme may be providing the good ones as .svg files

I am using mate-icon-theme for wayland and X11 session. With x11 everything is sharp with HIDIPI.
So i don't think it's an icon-theme problem.
But honestly i have no idea what going on here. It's not only the panel. My feeling is that whole the desktop and any application looks a bit blurred compare to X11.
Would be interesting how it looks with a mir wayland implementation.
Mate on wayland from Ubuntu @wmww  seems not get any update since 2021.
https://snapcraft.io/mate-wayland

lukefromdc:
Although I'm working with Wayland, Mir is another wlroots/foreign toplevel supporting wayland compositor from what I understand. It's my intention that this work should work with anything gtk-layer-shell can be used in, just as in x11 we can use marco, metacity, compiz, IceWM or almost anything else

lukefromdc:
Looks like we need to force the icons by adding
```
gtk-menu-images = true
gtk-button-images = true 
```
to the settings.ini file in each of our themes, 
users can still override this in ~/.config/gtk-3.0/settings.ini if desired. It works fine from the theme, and if added to the theme's settings.ini file works just fine. Only other way would be to have a settings manager running in wayland,  able to force the gsettings preference globally

I've run into an ugly problem with trying to force the icons to show in the applet menus in the code: The menus are loaded using
	```
gtk_action_group_add_actions (priv->panel_action_group,
	                              menu_entries,
	                              G_N_ELEMENTS (menu_entries),
	                              applet);
```
in `mate_panel_applet_init()` in mate-panel-applet.c 
and I don't know enough about Gtk Builder to pull a single GtkAction out of that so
`gtk_action_set_always_show_image` could be applied. No way to apply it to a whole gtk_action_group
so no easy way to do this. I only see gtk_image_menu_item and gtk_action having a function to force images (icons) to
render.

This icon issue effects every GTK application running under wayland, so we need a global fix anyway.

**Comment by lukefromdc** [comment]
Got a pleasant surprise testing fractional scaling in wayfire with caja and mate-panel: worked perfectly and didn't seem to bog things down. Also using different scaling on my 1080p secondary monitor and 4K main monitor has never been a problem in wayfire.

There may come a time when we can close out feature requests for fractional scaling with "use the wayland session." Scaling in x11 is a royal mess, it seems to "just plain work" in wayland.

**Comment by raveit65** [comment]
I could fix some things :)
After removing `outputs = kanshi` from wayfire.ini the scaling factor 1.7 is persistent.
During testing mate notifications in wayfire i got the error message about settings.ini in terminal.
`.config/gtk-3.0/settings.ini: Key file does not start with a group`
After adding the missing group `settings` to the file all icons are ther in menu.
```
[Settings]

gtk-menu-images = true
gtk-button-images = true
```
So it was my fault and really sorry about the confusion.

Btw. mate-notifications are working in wayfire when the daemon is running.
`/usr/libexec/mate-notification-daemon -r`

But the mate-notification-daemon won't be call automaticly with `notify-send.......`

**Comment by raveit65** [comment]
> Got a pleasant surprise testing fractional scaling in wayfire with caja and mate-panel: worked perfectly and didn't seem to bog things down. Also using different scaling on my 1080p secondary monitor and 4K main monitor has never been a problem in wayfire.
> 
> There may come a time when we can close out feature requests for fractional scaling with "use the wayland session." Scaling in x11 is a royal mess, it seems to "just plain work" in wayland.

Yes, fractional scaling is very nice

**Comment by lukefromdc** [comment]
Just tested mir, unfortunately current mir in the form of miral-system-compositor does NOT seem to support gtk-layer-shell, panel etc act as they do in Weston. Also tested Sway, which is strictly a tiling WM/compositor and thus quite different than Marco or Compiz.  

Big problem for a full session now is that Xwayland seems to be broken in Wayfire and in Sway as well, showing up as running in mate-system-monitor but not reachable by any application. Might be a wlroots issue, as wlroots has to support xwayland for it to work, and this might be partially broken.

Probably easier to find and fix this in wayfire or wlroots than it would be to write a whole new wlroots based compositor(which is quite beyond my current ability).

Wayfire was listed in a Gentoo wiki as one of the very few usable wayland compositors outside GNOME and KDE. The xwayland issue is actually the only problem I've had with it, but it's a big one with kdenlive, audacious, and GIMP all not supporting wayland yet. Likely to be one that partially goes away with time but will never be all the way gone as some legacy software will never be updated.

**Comment by Sunderland93** [comment]
> Just tested mir, unfortunately current mir in the form of miral-system-compositor does NOT seem to support gtk-layer-shell.

miral-shell --wayland-extensions wl_shell:xdg_wm_base:zwlr_foreign_toplevel_manager_v1:zwlr_layer_shell_v1 works well with mate-panel

**Comment by lukefromdc** [comment]
Ahh, that's the secret of how the Ubuntu mate-mir snap worked!  That will allow us both a basic compositor (in the role of marco) and a fancy one (in the role of compiz). Mir is being promoted for use in places like standalone kiosks, which implies it needs to be relatively light and hard for a user to screw up.

In my judgement, being able to run with more than one compositor is a major plus, one of the big plusses of GNOME 2 and then MATE after was the modular nature, where the WM, the file manager, and the panel were all interchangeable and all could be used elsewhere too

**Comment by raveit65** [comment]
> > Just tested mir, unfortunately current mir in the form of miral-system-compositor does NOT seem to support gtk-layer-shell.
> 
> `miral-shell --wayland-extensions wl_shell:xdg_wm_base:zwlr_foreign_toplevel_manager_v1:zwlr_layer_shell_v1` works well with mate-panel

Yeap, that works, but i have heavy graphical glitches when moving a window or with ubuntu's session start animation. I am using Ryzen 5000 internal amdgpu. With wayfire i don't get this. 
I added 
```
display-scale=2.0
display-config=single
startup-apps=mate-panel:caja
shell-terminal-emulator=mate-terminal
```
to ~/.config/.miral-shell.config.
Sadly, `startup-apps=mate-panel:caja --force-desktop -n` confused the options of miral-shell binary.

**Comment by lukefromdc** [comment]
I guess the next job is to start getting these applets buildable in or out of process?

**Comment by raveit65** [comment]
> I guess the next job is to start getting these applets buildable in or out of process?

That would be amazing when you find some time for it. I will support it with tests and reviews.

**Comment by lukefromdc** [comment]
BTW, turns out in sddm (installed as secondary display manager) I have the wayfire session listed. It comes up as my wayfire-MATE session save a couple bugs that may related to different session variable storage: Caja forgets desktop icon positions and view defaults, trash applet is on the panel but icon is invisible. None of this is a problem on wayfire from a VT command line.

**Comment by lukefromdc** [comment]
I've had a lot of trouble with autotools in trying to get out or in-process building optional for the wayland usable applets. I made it work for the very simple timer applet, but on both the trash and the battstat applet ran into ugly errors caused by autotools either not passing` ENABLE_IN_PROCESS` (or whatever else I tried) to the process of building the Makefile from Makefile.am or failing to read it, resulting in attempting to use both the code in the `ENABLE_IN_PROCESS` block and the` else` block at the same time. Resulting errors from that show attempts to specifiy building both with and without libtool.

Maybe as I work my way up the complexity chain from timer I will find the culprit?

For the timer, I just put the in-process Makefile.am text in the` ENABLE_IN_PROCESS` and the previous out of process code in the` else` block for now. On that applet it worked (after failing on two others) but still gave a warnign about NULL being defined twice, so I moved it to the top of the file:

`timerapplet/src/Makefile.am:41: warning: NULL was already defined in condition TRUE, which includes condition !ENABLE_IN_PROCESS ...
`
which may be a clue to the problems I've had with other applets, notably autotools reading Makefile.am as calling for building both with and without libtool no matter what I put in it, including using the in-process and out-of-process makefiles whole between if and else selectors

**Comment by cwendling** [comment]
I can help with the Autotools side of things if I know what's wanted.  I'll look at this probably this afternoon -- or maybe I'll still find time this morning, not sure.

**Comment by lukefromdc** [comment]
Thanks. What's needed is a way to on all of the applets except Accessx (only useful in X) and Geyes (cannot track pointer outside panel in wayland) to build in-process as a library in the selected libdir, or out of process as a standalone binary (as before) in libexec. I have a variable set in configure.ac (ENABLE_IN_PROCESS I think) and the timer applet can use it, but no matter what I tried with the Makefile.am files in battstat and trash I could not stop autotools from trying to use both sections of the relevent Makefile.am

Code that worked elsewhere failed there, with any variable I set somehow ignored

**Comment by cwendling** [comment]
> Thanks. What's needed is a way to on all of the applets except Accessx (only useful in X) and Geyes (cannot track pointer outside panel in wayland) to build in-process as a library in the selected libdir, or out of process as a standalone binary (as before) in libexec.

Done.  I also added support for AccessX and GEyes, as I think in/out of process is orthogonal to Wayland support, and having a `--enable-in-process` switch that doesn't make everything in-process felt weird.

I also took the liberty of pushing to this PR because it seemed simpler for everyone, but I can retire my commits if you want, and I can also squash some commits together if we're happy with the result (I'd rather keep some separated, like in/out process vs. Wayland, but that should be fairly easy).

> I have a variable set in configure.ac (ENABLE_IN_PROCESS I think) and the timer applet can use it, but no matter what I tried with the Makefile.am files in battstat and trash I could not stop autotools from trying to use both sections of the relevent Makefile.am Code that worked elsewhere failed there, with any variable I set somehow ignored

Ah yeah Autotools can be tricky.  I guess you encountered this? https://www.gnu.org/software/automake/manual/html_node/Objects-created-both-with-libtool-and-without.html
Unfortunately Automake is not smart enough to realize the conditionals make the problem irrelevant here, but the fix is easy enough (just add a target-specific `_CFLAGS`). And generating separate objects also makes it slightly more resilient if changing configuration options without cleaning the builddir, as Autotools isn't the best at rebuilding objects when only the environment flags changed.

**Comment by lukefromdc** [comment]
Thank you very much, I was stumped by this and really needed the help. Commiting these directly to the PR saves me some work too, will test now and look it over. I had written another version of configure.ac to allow selecting individual applets but set it aside for now due to the problems with "Objects-created-both-with-libtool-and-without"

**Comment by lukefromdc** [comment]
This make all the applets successfully build in-process or out. I got gdbus problems in out of process builds, probably an issue with the .service files, I should be able to fix that myself. It broke all but the AccessX applet at runtime but should be a simple fix.

Everything worked in-process, though AccessX won't be useful in wayland it works fine in process or out in X11. One or more of the applets generated these non-crashing errors, again should not be hard to fix. I should be able to take it from here. Thanks and thanks again for the assist!
```


(mate-panel:154145): GLib-CRITICAL **: 13:08:51.565: g_variant_ref: assertion 'value != NULL' failed

** (mate-panel:154145): WARNING **: 13:08:51.565: GDBus.Error:org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying

this non-crashing error in-process w all applets installed:

(mate-panel:238025): GLib-GObject-WARNING **: 13:18:03.684: g_object_connect: invalid signal spec "destroy"

(mate-panel:238025): Gtk-CRITICAL **: 13:18:03.690: gtk_widget_realize: assertion 'widget->priv->anchored || GTK_IS_INVISIBLE (widget)' failed

(mate-panel:238025): Gtk-CRITICAL **: 13:18:03.690: gtk_widget_realize: assertion 'widget->priv->anchored || GTK_IS_INVISIBLE (widget)' failed

(mate-panel:238025): Gtk-WARNING **: 13:18:04.337: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner WnckButton)

(mate-panel:238025): Gtk-WARNING **: 13:18:04.337: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner WnckButton)

```

**Comment by lukefromdc** [comment]
Looks like
```
(mate-panel:244415): Gtk-CRITICAL **: 13:29:31.160: gtk_widget_realize: assertion 'widget->priv->anchored || GTK_IS_INVISIBLE (widget)' failed

(mate-panel:244415): Gtk-CRITICAL **: 13:29:31.160: gtk_widget_realize: assertion 'widget->priv->anchored || GTK_IS_INVISIBLE (widget)' failed

```
in-process is geyes, which also has issues in-process correctly following the pointer unless put on the left side of the screen. I will need to add code to get the applet position along the panel to fix that.

**Comment by lukefromdc** [comment]
Trash out of process didn't work because of a typo in /data/Makefile.am
"libexecir" instead of "libexecdir" easy fix, pushing now. Nothing to worry about, this happens to all of us

**Comment by lukefromdc** [comment]
All builds now finish and run out of process, now to fix some in-process errors that cropped up

**Comment by lukefromdc** [comment]
Now everything basically works in-process too. Couple runtime warnings to deal with, and if possible get geyes to properly follow the pointer when out of process. If not we may have to be able to build it out of process with everything else in-process as an option, probably specified as --enable-geyes-in-process (experimental)

**Comment by lukefromdc** [comment]
There is an issue (at least in some themes)  in AccessX out of process showing a rotated background on a panel with an image background if the panel is changed from vertical to horizontal or horizontal to vertical. This is inherited from master and not caused by any of these commits-and is fixed by building in-process. An in-process build eliminates the GtkPlug window on which the background is redrawn.

**Comment by lukefromdc** [comment]
We are now rid of the runtime warning from geyes, but it still does not properly track the pointer when built in-process. Note that in wayland it cannot track the pointer when out of the panel due to wayland restrictions on one app watching another app's windows

**Comment by lukefromdc** [comment]
Last force-push makes geyes track the pointer correctly when in-process by compensating for applet position on the panel, no change in out of process behavior

File: trashapplet/data/Makefile.am [hunk 2]
```diff
@@ -4,7 +4,7 @@ service_in_files = org.mate.panel.applet.TrashAppletFactory.service.in
 if ENABLE_IN_PROCESS
 APPLET_LOCATION = $(pkglibdir)/libmate-trash-applet.so
 else !ENABLE_IN_PROCESS
-APPLET_LOCATION = $(libexecir)/trashapplet
+APPLET_LOCATION = $(libexecdir)/trashapplet
```
cwendling, line 7: Oops, good catch

netspeed/src/Makefile.am
cwendling, line 1: Actually `BUILT_SOURCES` works, but I had a typo when using the variable, oops again.  Fixed in 3b9c0b9504a811733ed606ce69bcdab0fa832e98.

It's better to use [`BUILT_SOURCES`](https://www.gnu.org/software/automake/manual/html_node/Sources.html) than putting generated sources in the regular ones to solve some cases of bootstrapping dependencies between objects and included files.

accessx-status/src/applet.c
```diff
@@ -1634,11 +1634,18 @@ accessx_status_applet_fill (MatePanelApplet* applet)
...
+                      G_CALLBACK (accessx_status_applet_destroy),
+                      sapplet);
+
+    g_signal_connect (sapplet->applet, "change-orient",
+                      G_CALLBACK (accessx_status_applet_reorient),
+                      sapplet);
+
+    g_signal_connect (sapplet->applet, "change-size",
+                      G_CALLBACK (accessx_status_applet_resize),
+                      sapplet);
```
cwendling, line 1648: Actually this should not have anything to do with in or out of process, and is actually a regression from dc37d79d6e2797f7f613051261466b93fc01107d. [`g_object_connect()`](https://docs.gtk.org/gobject/method.Object.connect.html) actually expects a prefix to know how to connect the signal.  Here, what has been lost in dc37d79d6e2797f7f613051261466b93fc01107d is the `signal::` prefix.

accessx-status/src/applet.c
```diff
@@ -1632,11 +1634,18 @@ accessx_status_applet_fill (MatePanelApplet* applet)
...
```
cwendling, line 1648:
```suggestion
    g_object_connect (sapplet->applet,
                      "signal::destroy", accessx_status_applet_destroy, sapplet,
                      "signal::change-orient", accessx_status_applet_reorient, sapplet,
                      "signal::change-size", accessx_status_applet_resize, sapplet,
                      NULL);
```

cwendling: Either way if fine though, using the convenience only saves a bit of typing and duplication.

lukefromdc: Testing this now, I can see what you did different using` "signal::<something>"`

lukefromdc: Works fine both in-process and out, will commit this

accessx-status/src/applet.c
lukefromdc: Your proposed fix worked, committed it and about to push

lukefromdc:
Oh yeah this will need squashing. A per-applet squash won't be possible on making in-process builds possible directly, and such work absolutely requires a backup copy of the rep. Would be very easy to break it

lukefromdc:
Commit history squashed to something easier to work with.

raveit65:
My actual testing results:
Multiload applet doesn't display anymore in x11 and wayland.
brightness.applet - doesn't display under wayland
dictionary -  doesn't display under wayland
trash-applet - no right-click menu under wayland, instead you will get the panel menu.

Those applets are there in add-to-pannel gui.

raveit65:
When building applets out-of-process the multiload applet displays well under X11.

lukefromdc:
I fixed multiload in the last force-push, one line needed in  it's data/Makefile.am (the one to write true or false for "in-process=" hadn't made it in with the transition to selectable in or out of process builfs

I will have to re-check trash to ensure my fix for clicking on it in wayland didn't get lost, i can easily bring it back if it did

lukefromdc:
Ahh  I see it now-the trashapplet's  right-click context menu was lost in and only in wayland. Should be easy to fix

raveit65:
Ok, multiload do  run after a rebuild. But trash-applet doesn't have a menu
Also brightness and dictionary do not run under wayland.
Not sure if they were OK in my first test.

lukefromdc:
In trashapplet, it seems there's an issue with clicks being "eaten." In wayland, opening trash on button release doesn't work, the panel eats the click and worse yet responds by moving the panel to the top if it's on the bottom. Opening on button press works fine with left click. That however then causes the right click to be ignored, possibly the original reason for using left button release to open the trash. Thus we will need an explicit command to open the right click menu on button 3 in 
`trash_applet_button_release` which remains the function name at this time

lukefromdc:
Using
`  if ((event->button == 1) || (event->button == 3)) `` in trash_applet_button_release` will open the trash on left or right click, proving we still have access to both clicks. Just need to redirect button 3 to opening the context menu now

raveit65:
> Using `` if ((event->button == 1) || (event->button == 3)) `` in trash_applet_button_release `` will open the trash on left or right click, proving we still have access to both clicks. Just need to redirect button 3 to opening the context menu now

I recall that this was working in wayland during my first test.

lukefromdc:
Got it working with code from cpufreq-applet. Seems we had to exclude some of the other cases. At first the right click worked but not the left click, then the reverse of that but never had a problem w this from cpufreq-applet so borrowed that code

lukefromdc:
Squashed all the cleanups too

raveit65:
I just sent a mail to travis-ci for getting new credits again.

lukefromdc:
Thanks

lukefromdc:
Brightness and inhibit are provided by mate-power-manager, dictionary by mate-utils. Dictionary can certainly be ported in that package, brightness and inhibit may be using Xorg-specific functions but I will have to check.

raveit65:
Ahh ok, those applets are from different packages.
I think also the kill-process applet use X functions to find the windows and should be disable for wayland.

lukefromdc:
force-quit is provided by mate-panel and indeed we should disable it in wayland

lukefromdc:
Speaking of applets provided by mate-panel, connect to server works well, drawer works but renders at the start of the panel so needs code to move it to the applet. Separator and search for files work fine in wayland. Do not disturb does not work (does it work by creating a status icon though?), logout works but right now doesn't have a session manager to connect to so does nothing.  Shut down same behavior as logout (nothing to connect to) if present after adding in an Xorg session but for some reason is not shown in the add to panel dialog in wayland. 

All of these are applets provided by mate-panel, I will have to iterate through the ones with issues

raveit65:
My actual results in wayfire session:

battstat loads fine, menus are there.
charpick loads fine, menus are there.
command loads fine, menus are there.
CPUfreq loads fine, menus are there, cpu governor can't be selected, not related to this PR.
Drivemount loads fine, menus are there.
weather, loads fine, menus are there, but has problems to load selected location, not related to this PR, ihmo.
multiload , loads fine, menus are there.
netspeed , loads fine, menus are there.
sticky notes ok, but gives runtime warning when removing the applet from the panel.
```
(mate-panel:195905): Wnck-WARNING **: 20:15:09.628: libwnck is designed to work in X11 only, no valid display found

(mate-panel:195905): Wnck-CRITICAL **: 20:15:09.629: wnck_screen_force_update: assertion 'WNCK_IS_SCREEN (screen)' failed
```
timer-applet OK but runtime warning when timer ends, not related to PR
```
(mate-panel:196085): Atk-CRITICAL **: 20:19:09.594: atk_object_set_description: assertion 'description != NULL' failed
```
trash-applet , loads fine, menus are there.

I think the warnings from sticky-notes-applet are from mate-panel code, but we should investigate why this happens.
All other applets are fine now.

I will do a code-format review the next days.

raveit65:
> Do not disturb does not work (does it work by creating a status icon though?)

I think it is out-of-process.
From my spec file of mate-notification-daemon
%{_datadir}/dbus-1/services/org.mate.panel.applet.MateNotificationAppletFactory.service

lukefromdc:
I will look into the stickynotes warning, usually that's signals not getting disconnected. Warnings are invisible in terminal for out of process so these may or may not be new

lukefromdc:
Last force-push fixed the stickynotes warnings, by limiting some x11 and wnck stuff to x11 only. Note that in wayland there is no easy way to exclude the note windows from the window list, and they are decorated. It would probably require a new function in libmate-panel-applet to let applets talk to the window list, and stickynotes would initially be the only user of it.

The decorated windows could be fixed and displaying notes on all four windows enabled by using gtk-layer-shell, but this would require adding that as a new conditional dependency for the entire package, plus another config option to enable wayland support for sticky notes just to undecorate them and allow showing notes on all desktops.

Also note that hiding notes (whether by clicking on the desktop or from the menu item) didn't work the last time I tested sticky notes while testing another PR quite a while ago. Doesn't in wayland but doesn't in x11 either so same behavior there and probably same as current git master in x11.

raveit65:
> Last force-push fixed the stickynotes warnings, by limiting some x11 and wnck stuff to x11 only.

Confirmed, warning is gone, thanks.

IHMO, all applets  should load/unload to panel and preferences windows should be open/close nicely with this PR.
Specific applet functions which do not work at this moment can be fix in single PRs later.
Other things like session-management, m-s-d and open caja issues are more important.

Battstat gives a warning when removing from panel, but for me this can be fix later.
1 open battstatt preferences
2 close preferences
3 remove battstat from panel
```
(mate-panel:96778): Gtk-CRITICAL **: 12:13:29.975: gtk_widget_destroy: assertion 'GTK_IS_WIDGET (widget)' failed
```
All other applets load and unload fine under wayland.

@cwendling @lukefromdc 
With the interactive rebase command it's possible to split the cleanup commits into single commits for every applet, which can be squashed in the main commit for every applet.
Ie. for last commit:
1. choose edit in vi rebase terminal. Now all changes are uncommented and you can add them again to git staging area.
2. git add accessx-status/src/applet.c
3. git commit -a -m "accessx cleanup"

Now the same procedure for trashapplet/src/trashapplet.c
and at least for the remaining sticky-notes changes.
I would do this splitting first for the first cleanup comment to avoid conflicts/problems. https://github.com/mate-desktop/mate-applets/pull/651/commits/d9e90ff16a0d4c0f4a8e2f7f7a94cc53e8dd1ed1

I can do that for you if you like.

lukefromdc:
I have never attempted to split up a commit. If you can handle that it's greatly appreciated

raveit65:
Done, i split up all commits with changes from files from different applets and sorted the history. The final squashing i leave for you to avoid that all commits get my name. I hope i didn't break anything. Please review.
In a test branch i squashed the commits and CI building was fine.
https://github.com/mate-desktop/mate-applets/commits/in-process-test-commit-squashed

lukefromdc:
Testing now, both ways in x11 plus in-process on wayland

lukefromdc:
Everything seemd to work in the test branch: all applets out of process x11, in-process x11, and in-process wayland where the dialog came up to bypass applets flagged as not supported in wayland. Even a test stickynote created on an x11 desktop came up on the wayfire desktop as well.  Running `mate-panel --replace` did not bring up any warnings in terminal

geyes/src/geyes.c
```diff
@@ -144,9 +145,17 @@ timer_cb (EyesApplet *eyes_applet)
             gdk_window_get_device_position (gtk_widget_get_window (eyes_applet->eyes[i]),
                                             gdk_seat_get_pointer (seat),
                                             &x, &y, NULL);
+
+#ifdef ENABLE_IN_PROCESS
+            gdk_window_get_origin (gtk_widget_get_window (GTK_WIDGET(eyes_applet->applet)),
+                                       &applet_x, &applet_y);
```
raveit65, line 151: Indent

lukefromdc:
https://github.com/mate-desktop/mate-applets/tree/in-process-squashed-to-one
Is the same commits squashed into one with a well-commented commit message preserving the history of how they came to be. Test builds coming in minutes

stickynotes/stickynotes.c
```diff
@@ -823,14 +821,16 @@ stickynotes_save_now (void)
 
     xmlDocSetRootElement (doc, root);
     xmlNewProp (root, XML_CHAR ("version"), XML_CHAR (VERSION));
-
-    wnck_screen = wnck_screen_get_default ();
-    wnck_screen_force_update (wnck_screen);
-
+#ifdef GDK_WINDOWING_X11
+    GdkDisplay *display = gdk_screen_get_display(gdk_screen_get_default());
+    if (GDK_IS_X11_DISPLAY(display))
```
raveit65, line 826:
```suggestion
    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
    if (GDK_IS_X11_DISPLAY (display))
```

lukefromdc: Easy to handle, can do it now

stickynotes/stickynotes.c
```diff
@@ -948,10 +951,12 @@ stickynotes_load (GdkScreen *screen)
     xmlDocPtr doc = NULL;
     xmlNodePtr root;
     xmlNodePtr node;
-    /* WnckScreen *wnck_screen; */
     GList *new_notes, *tmp1;  /* Lists of StickyNote*'s */
     GList *new_nodes; /* Lists of xmlNodePtr's */
     int x, y, w, h;
+#ifdef GDK_WINDOWING_X11
+    GdkDisplay *display = gdk_screen_get_display(gdk_screen_get_default());
```
raveit65, line 958:
```suggestion
    GdkDisplay *display = gdk_screen_get_display (gdk_screen_get_default());
```

lukefromdc: Done

lukefromdc: Ahh, more of them, will get them now and push again

stickynotes/stickynotes.c
lukefromdc: Got that one too, it's in squashed-to-one

geyes/src/geyes.c
lukefromdc: If you look at the lines above, this block looks like, but is NOT inside the ` if` block that is in turn inside a `for` block, that's what fooled me. It is inside the` for` block only, fixing this now

lukefromdc: Correction posted above. Original bracket location hard to see so moving it

lukefromdc: Pushed bracket change to in-process-squashed-to-one as below:

```
    for (i = 0; i < eyes_applet->num_eyes; i++)
    {
        if (gtk_widget_get_realized (eyes_applet->eyes[i]))
        {
            gdk_window_get_device_position (gtk_widget_get_window (eyes_applet->eyes[i]),
                                            gdk_seat_get_pointer (seat),
                                            &x, &y, NULL);

#ifdef ENABLE_IN_PROCESS
            gdk_window_get_origin (gtk_widget_get_window (GTK_WIDGET(eyes_applet->applet)),
                                       &applet_x, &applet_y);
#else
            applet_x = 0;
            applet_y = 0;
#endif
            gtk_widget_translate_coordinates (eyes_applet->eyes[i],
                                              gtk_widget_get_toplevel(eyes_applet->eyes[i]),
                                              -applet_x, -applet_y, &dx, &dy);
            x -= dx;
            y -= dy;

            if ((x != eyes_applet->pointer_last_x[i]) ||
                (y != eyes_applet->pointer_last_y[i])) {

                calculate_pupil_xy (eyes_applet, x, y, &pupil_x, &pupil_y,
                                    eyes_applet->eyes[i]);
                draw_eye (eyes_applet, i, pupil_x, pupil_y);

                eyes_applet->pointer_last_x[i] = x;
                eyes_applet->pointer_last_y[i] = y;
            }
        }
    }
    return TRUE;
```

lukefromdc: This should be easier to read and less likely to trip up anyone working on it later

stickynotes/stickynotes.c
lukefromdc: done

lukefromdc:
https://github.com/mate-desktop/mate-applets/commit/675ee7474fedde6d0c28d60e9ddbd19587bf1634
is the final single commit (barring further edits) in https://github.com/mate-desktop/mate-applets/tree/in-process-squashed-to-one  that can replace everything in the in-process branch via a rename and force-push for final merge, thus keeping it all connected to this history

raveit65:
Really, you squashed all commits in one.....!!!
Sorry, i spend 2 hours of my life time to split commits of your dirty and confused work.
This is really a killing joke and sucks.
.... or your skills are sufficient  to use the interactive rebase command.
I am out.......sigh

lukefromdc:
I kept a backup, so if you want one commit per applet I can do that too almost as easily, no problem

lukefromdc:
I can rebase easily enough, splitting commits was something I have never done and would have to do manually w files

lukefromdc:
Will switch this (again from your work) to one per applet. That's why I kept a backup

lukefromdc:
Done: one commit for each applet, and the final commit with the option to build in-process or out, plus the cleanups. Again I have backups if needed

lukefromdc:
It was the words " The final squashing i leave for you to avoid that all commits get my name. " that confused me. I thought you wanted all in one with a composite commit message from that line. Sorry about that but it's fixed now

lukefromdc:
Will test again, thanks

raveit65:
I restored your name from cpufreq commit,
and hopefully you understand that one huge massive commit isn't helpful for debuging.

lukefromdc:
No problem at all, I just misread your ask. I am autistic and can easily make that sort of mistake

lukefromdc:
Just got successful tests of in-process and out of process builds, in-process on both x11 and wayland. On the timer I found a behavioral difference: in x11 you can left click the appled to pause or restart the timer, in wayland you cannot. Probably same fix as from trash for that. I can deal with that now or after this is merged, whichever you prefer. It is wayland specific rather than in/out process specific

lukefromdc:
This patch (note the filenames specific to my filesystem) fixes the timer applet behavior in wayland. Add it here or hold it for a separate, wayland-specific PR?
```
--- /home/luke/Desktop/Development/MATE_Development_Work/AA_Master/mate-applets/timerapplet/src/timerapplet.c	2023-07-16 18:43:08.357224055 -0400
+++ /home/luke/Desktop/Timer_Applet_Fix/timerapplet.c	2023-07-16 21:45:49.545053966 -0400
@@ -27,6 +27,7 @@
 #include <glib.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
+#include <gdk/gdkx.h>
 #include <gtk/gtk.h>
 
 #include <libnotify/notify.h>
@@ -353,14 +354,17 @@ timer_preferences_callback (GtkAction *a
 }
 
 static gboolean
-timer_applet_click (TimerApplet *applet)
+timer_applet_click (TimerApplet *applet,  GdkEventButton *event, gpointer userdata)
 {
-    if (!applet->active && !applet->pause && applet->elapsed)
-        timer_reset_callback (NULL, applet);
-    else if (applet->active && !applet->pause)
-        timer_pause_callback (NULL, applet);
-    else if (!applet->active || applet->pause)
-        timer_start_callback (NULL, applet);
+    if (  event->button == 1)
+    {
+        if (!applet->active && !applet->pause && applet->elapsed)
+            timer_reset_callback (NULL, applet);
+        else if (applet->active && !applet->pause)
+            timer_pause_callback (NULL, applet);
+        else if (!applet->active || applet->pause)
+            timer_start_callback (NULL, applet);
+    }
     return FALSE;
 }
 
@@ -423,10 +427,16 @@ timer_applet_fill (MatePanelApplet* appl
     g_signal_connect (applet->applet, "destroy",
                       applet);
-
+#ifdef GDK_WINDOWING_X11
+  if (GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
     g_signal_connect_swapped (applet->applet, "button-release-event",
                               G_CALLBACK (timer_applet_click),
                               applet);
+  else
+#endif
+    g_signal_connect_swapped (applet->applet, "button_press_event",
+                              G_CALLBACK (timer_applet_click),
+                              applet);
 
     /* set up context menu */
     applet->action_group = gtk_action_group_new ("Timer Applet Actions");
```

raveit65:
I would do it in a separated PR.
It's not related to in/out-of process thematic.
I will do a last test and will approve PR if all works well.

lukefromdc:
OK, that's what I will do. Holding it until this one is done as it depends on this.

raveit65:
Ok, i can confirm that actual state of this branch is the same like my local copy before squashing.
Can please squash branch that we have only one commit for each applet. 
This should be better for debuging.

lukefromdc:
I can do that by and only by making a new branch from master, moving across configure.ac and committing it, then doing the same for every applet. I will leave that as a separate branch for now, as I do not know what you want to do after

I find VI very difficult to use, no way I can split up commits in it at this point. I can get the same job done by different means, files w

raveit65:
Wait Wait, for me it is OK now. We one single commit for each applet and one for general configure things.
No need to spend more time for it.

raveit65 (approved):
LGTM

lukefromdc:
Timer fix PR is at
https://github.com/mate-desktop/mate-applets/pull/654

lukefromdc:
The battstat warning on opening and then closing preferences, then removing from the panel did not happen here on x11 nor on wayland. Wonder what is different?

raveit65:
> The battstat warning on opening and then closing preferences, then removing from the panel did not happen here on x11 nor on wayland. Wonder what is different?

I can't reproduce it anymore on wayland. Seems  it was a silent self healing issue :-) 
Ignore it.

PR #654, opened 2023-07-18
Timer-catch the left click when under wayland

lukefromdc:
Ensure the timer applet can be clicked on with the left mouse button in wayland. Otherwise the timer cannot be started and stopped by clicking on the applet in wayland. Left-click goes through to the panel instead without this

### DIFF
--- a/timerapplet/src/timerapplet.c
+++ b/timerapplet/src/timerapplet.c
@@ -27,6 +27,7 @@
 #include <glib.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
+#include <gdk/gdkx.h>
 #include <gtk/gtk.h>
 
 #include <libnotify/notify.h>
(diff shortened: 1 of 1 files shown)


timerapplet/src/timerapplet.c
```diff
@@ -423,10 +429,16 @@ timer_applet_fill (MatePanelApplet* applet_widget)
     g_signal_connect (applet->applet, "destroy",
                       G_CALLBACK (timer_applet_destroy),
                       applet);
-
```
raveit65, line 426: This blank line shouldn't be removed.

raveit65:
PR does what it promised. Only the atk runtime warning is still there under wayland but this is unrelated to PR .

lukefromdc:
I'll add back the blank line, and in another PR fix that ATK warning.

timerapplet/src/timerapplet.c
lukefromdc: Brought the missing blank line back in latest force-push, thanks

raveit65 (approved):
LGTM

PR #655, opened 2023-07-18
timer: fix in-process ATK warning

lukefromdc:
Fix `atk_object_set_description: assertion 'description != NULL' failed` warning when timer runs down. Looks like a race condition is present here. Ideally someone with a working orca install (errors out here) would test this, but the warning indicates that we would not be setting anything when` atk_object_set_description` gets called with an empty description anyway.

Alternately, we could set that description to "finished timer" but Orca should already speak the label text, which shows up in Accerciser

### DIFF
--- a/timerapplet/src/timerapplet.c
+++ b/timerapplet/src/timerapplet.c
@@ -162,7 +162,7 @@ timer_callback (TimerApplet *applet)
             gtk_widget_set_tooltip_text (GTK_WIDGET (applet->label), name);
             gtk_widget_hide (GTK_WIDGET (applet->pause_image));
             atk_object_set_name (atk_obj, label);
-            atk_object_set_description (atk_obj, tooltip);
+            atk_object_set_description (atk_obj, "");
 
             if (g_settings_get_boolean (applet->settings, SHOW_NOTIFICATION_KEY))
             {


raveit65:
I can confirm that PR fixes the warning when build in-process and running in wayfire session.
No idea if this is a valid fix.

lukefromdc:
Should we hold this pending an A/B/A test with a working Orca install and knowing what's supposed to be spoken there? I could not figure out with Accerciser what's supposed to be going on

raveit65:
I think @cwendling can help as member of hypra.fr 
It' only a warning and we don't have any rush, ihmo.

lukefromdc:
At least Orca is supposed to work on wayland as well as x11, though git master Orca is needed as of the last report on the GNOME wiki:

https://wiki.gnome.org/Accessibility/Wayland

Thus by the time distrops are shipping a mate-wayland session, Orca should work with it mostly the same as it does now in x11.

lukefromdc:
What I can reasonably figure about this is the warning is aboutt setting an empty description, which would never be useful. Whether this is happening because of a race condition (e.g trying to read the tooltip text while it is changing), because we should not read (and Orca should not speak) the tooltip text when the tooltip is not showing, or something else I do not know.]

Using the tooltip text when we are setting a description unconditionally strikes me as odd. If we want Orca to always have access to that, the text should go to ATK and the tooltip in parallel and not in series. I do NOT know if that is intended though.

This would depend on whether Orca picks up the dialog or notification and speaks it or not, and with Orca broken here (throws a python error), I cannot test that.

zhuyaliang (commented):
Would it be better to use ```gtk_widget_get_tooltip_text``` gets the contents of the tooltip for widge??

lukefromdc:
I can test that and find out.

lukefromdc:
Found the problem: ` tooltip` is a `gchar` not the tooltip itself. The
`if (remaining <= 0)`
 block in question is the block covering the case of the timer having counted down to zero or not having been started. In this block,` tooltip` is set to NULL and is not touched again until we attempt to read it into atk_object_set_description(). 

https://docs.gtk.org/atk/method.Object.set_description.html
Using NULL for description is not allowed, for an empty value we are supposed to use `"" `instead. 
When the timer is running, the
`else`
block runs, and the value of` tooltip` is set to the remaining time. The correct fix is not to put this in a conditional (which leaves `tooltip` set to the last non-zero value) but rather to set it to` ""` in the no time remaining block. Will try it now and force-push if it works

timerapplet/src/timerapplet.c
```diff
@@ -162,7 +162,9 @@ timer_callback (TimerApplet *applet)
             gtk_widget_set_tooltip_text (GTK_WIDGET (applet->label), name);
             gtk_widget_hide (GTK_WIDGET (applet->pause_image));
             atk_object_set_name (atk_obj, label);
-            atk_object_set_description (atk_obj, tooltip);
+
```
zhuyaliang, line 165: Excess blank lines

lukefromdc: I can get those out now, in another PR have to kill a build warning from earlier work

lukefromdc: Also found the code flow was different than I thought (`tooltip `never takes the time remaining) but the fix is the same

lukefromdc: Done, excess blank lines are gone

zhuyaliang (approved):
Warning fixed

PR #659, opened 2023-08-14
trashapplet: Remove animation that hasn't worked in years

lukefromdc:
*The trash applet hasn't had working animation in many years
    *xstuff_zoom_animate also appears in mate-panel with 4 arguments instead of 2.
    *conflicting function names caused a segfault when built in-process


lukefromdc:
Previously we fixed one segfault (in trash) by excluding the conflicting function from in-process builds of the trash applet. With this we remove any potential segfaults caused by any other applet of the panel calling the wrong function. Why the conflicting functions ever worked I do not know, wondering if the  older glib version somehow allowed the applet's version of xstuff_zoom_animate() instead of the panel's function of the same name to be called first.

This was not a glib bug, but rather a bug in our code revealed by a glib update, as it should never have worked and it it did was working by chance only.

muktupavels:
This will work but if you want avoid similar problems in future then read documentation for `g_module_open`.

P.S. You don't need to change name of static function.

raveit65:
The crash is gone with using the internal zoom-animation.
The animation itself do not work, same when building out-of-process.
But this is another issue not related to PR.

lukefromdc:
The question is do we want to keep the currently non-working  animation around in case someone comes up with a fix for it later, or remove it as it hasn't worked in years. The latter would allow removal of xstuff.c and xstuff.h from the applet's source files, but would complicate any later rewrite aimed at making the trash icon respond to being clicked on.

raveit65:
+1 for removing. I don't think that someone will fix it because nobody knows that the function exists.
And is easy to restore with a git revert.

lukefromdc:
OK, that's what I will do

raveit65 (approved):
LGTM

PR #660, opened 2023-10-06
replace deprecated gtk_image_menu_item*

raveit65:


zhuyaliang (approved):
Compilation warning removed

lukefromdc (approved):
Builds and runs fine, and does not throw warnings in terminal when running the panel from terminal (in-process build tested which would cause any runtime warnings to be displayed in terminal).

Issue #668, opened 2024-04-02
After update to 1.28 Sticky Notes stopped working


kotoyants:
#### Expected behaviour
New note windows when double-clicking icon or using menu

#### Actual behaviour
Nothing happens, error in journal
`stickynotes-app[227298]: gtk_window_set_screen: assertion 'GTK_IS_WINDOW (window)' failed
stickynotes-app[227298]: gtk_window_set_decorated: assertion 'GTK_IS_WINDOW (window)' failed
stickynotes-app[227298]: gtk_window_set_skip_taskbar_hint: assertion 'GTK_IS_WINDOW (window)' failed
stickynotes-app[227298]: gtk_window_set_skip_pager_hint: assertion 'GTK_IS_WINDOW (window)' failed
stickynotes-app[227298]: gtk_widget_add_events: assertion 'GTK_IS_WIDGET (widget)' failed
stickynotes-app[227298]: gtk_widget_add_events: assertion 'GTK_IS_WIDGET (widget)' failed
stickynotes-app[227298]: gtk_text_view_get_buffer: assertion 'GTK_IS_TEXT_VIEW (text_view)' failed
stickynotes-app[227298]: gtk_widget_add_events: assertion 'GTK_IS_WIDGET (widget)' failed
stickynotes-app[227298]: gtk_widget_add_events: assertion 'GTK_IS_WIDGET (widget)' failed
stickynotes-app[227298]: gtk_widget_add_events: assertion 'GTK_IS_WIDGET (widget)' failed
stickynotes-app[227298]: gtk_widget_set_direction: assertion 'GTK_IS_WIDGET (widget)' failed
stickynotes-app[227298]: gtk_window_stick: assertion 'GTK_IS_WINDOW (window)' failed
stickynotes-app[227298]: gtk_window_resize: assertion 'GTK_IS_WINDOW (window)' failed
stickynotes-app[227298]: gtk_image_set_from_icon_name: assertion 'GTK_IS_IMAGE (image)' failed
stickynotes-app[227298]: gtk_image_set_pixel_size: assertion 'GTK_IS_IMAGE (image)' failed
stickynotes-app[227298]: gtk_image_set_from_icon_name: assertion 'GTK_IS_IMAGE (image)' failed
stickynotes-app[227298]: gtk_image_set_pixel_size: assertion 'GTK_IS_IMAGE (image)' failed
[...]

PR #671, opened 2024-11-22
Fix stickynotes

lukefromdc:
Fix failure of stickynotes to show in both x11 and wayland, and wayland segfaults. Do not use GtkSourceView anymore since the current version is not working with stickynotes, use GtkTextView which works fine even if it has fewer features.

Also ensure that when in Wayland (or any future non-x11 windowing system) we don't call x11 specific functions

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -297,15 +297,8 @@ AC_ARG_ENABLE([stickynotes],
     AS_HELP_STRING([--enable-stickynotes], [Enable stickynotes applet.]),
     enable_stickynotes=$enableval,
     enable_stickynotes=yes)
-if test "x$enable_stickynotes" = "xyes"; then
-    PKG_CHECK_MODULES(STICKYNOTES, gtksourceview-4,
-                      have_gtksourceview=yes, have_gtksourceview=no)
 
-    if test "x$enable_stickynotes" = "xyes" -a "x$have_gtksourceview" = "xno"; then
-        AC_MSG_ERROR([Stickynotes explicitly requested but gtksourceview not found])
-    fi
-fi
-AM_CONDITIONAL(BUILD_STICKYNOTES_APPLET, test "x$have_gtksourceview" = "xyes")
+AM_CONDITIONAL(BUILD_STICKYNOTES_APPLET, test "x$enable_stickynotes" = "xyes")
 
 dnl ***************************************************************************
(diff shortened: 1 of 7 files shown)


lukefromdc:
Possible fix for
https://github.com/mate-desktop/mate-applets/issues/668

lukefromdc:
Once again, we've got travis problems different from before and in at least four attempts I have not been able to fix them. Giving up on that for tonight. After my last try curl wasn't found for some reason  Same in master though, will force-push just to keep this synched

robertkirkman:
Thank you very much this worked to fix the problem for me! "New Note" now works, and the applet does not crash anymore.

lukefromdc:
Just to help with getting the broadest range of testing, do you build mate-applets with the applets out of process(the default) or (newly supported for wayland compatabiity)  in-process? I tested an in-process build compatable with wayland, in both x11 and in wayland and it worked both ways. If you build out of process, we now have verification this works both ways. In that case I will wait a bit to see if any other team member wants to review it, and if not will merge it,

Distros probably won't pick it up until a new release though, and that requires fixing (again) the travis problems. I don't have any other way to push packages made by ` make distcheck` to https://pub.mate-desktop.org/

robertkirkman:
I use Xorg, so I guess that means they are out-of-process for me (it is named  `/usr/libexec/mate-applets/stickynotes-applet` for me so I think that's what that means?)

lukefromdc:
Confirm, that means you built them out of process.That works only in Xorg, MATE's default, and the PR worked for you in that build. In-process is compatable with both Xorg and Wayland and that build works for me. That verifies I didn't create a fix that only works one way.

lukefromdc:
I have merged this and attempted to fix the Travis issues, but no success on the release. 

Travis is nearly impossible for me to maintain the use of without assistance, and is not something I worked on most of my time here on MATE. Travis is changing fast now and seems to constantly break, blocking releases. 

I need either help working with Travis or a way to stop using it and push releases to https://pub.mate-desktop.org/releases/ from Github. I really don't know much about remote work in general though.

lukefromdc:
1.28.1 containing this fix is now available, distros offering 1.28 should pick it up soon
https://pub.mate-desktop.org/releases/1.28/mate-applets-1.28.1.tar.xz

Issue #674, opened 2025-01-17
[Mate 1.28] geyes no longer seems to work properly if in bottom panel or right panel


JamesR98:
The bottom panel seems to think anytime the cursor is lower than half the height of the screen the eyes should be looking down (when the cursor is clearly above the eyes). The right panel seems to think it should always be looking right.

The left and top panel seems to work.

I've tried changing themes and adjusting the height of panels as that sometimes has an effect on applets, but behavior was consistent across all my tests.

I suspect it has to do with this change:
https://github.com/mate-desktop/mate-applets/pull/621/commits/70216007b609d586fd4a8e1f17b60555f582119a
but I'm not sure.

lukefromdc:
Did this ever work on a left or right panel? I don't know if this was ever written to support left and right panels, a lot of applets got work in the 2015-2020 era to improve vertical panel support in many applets. This particularily benefited Ubuntu-MATE's optional "mutiny" layout that resembled the Unity desktop with a wide left panel and a slim top panel

JamesR98:
> Did this ever work on a left or right panel?

geyes 1.28 works just fine in the left panel, only the right and bottom panels do not work.

On version 1.26.1 the eyes work on all 4 panels.

lukefromdc:
Try reverting the suspect commit and see what happens.

JamesR98:
Sorry I'm not much of a developer so I won't really be able to help  that much, I have it built however Fedora 41 builds it :/

lukefromdc:
If you can open the package in Engrampa, look to see whether you have a file named geyes.so that goes into /usr/lib or just one named geyes that goes into /usr/libexec (or in both cases whereever Fedora puts them). If you have .so files for geyes (and everything else) you have mate-applets build in-process, meaning it can be used with wayland as well as with x11. If you have the bare applet names for these files (no .so extension) they are built out of process as they always were until 1.28.

This matters because the code for finding the pointer is different, and if you have mate-applets built in-process you may have found a bug in code I wrote. I have my normal panel on the bottom, normally test applets I don't normally use on a top panel, so this (and also side panel issues) could have easily gotten by me.

To test any change I will of course have to close down my usual mate-wayland session and fire up an x11 session. Geyes cannot work in wayland because wayland won't let the applet see where the pointer is unless it is inside the applet's window.

JamesR98:
Here is are the eyes files that aren't in /usr/share/:
```
[...]

PR #675, opened 2025-01-23
Geyes in-process: follow the mouse properly no matter where the apple…

lukefromdc:
*In-process: fix bottom panel/side panel issue where eyes fail to properly track the mouse.

*The original out of process code for finding the applet position seems to work fine in-process
*When in-process, correct for position of individual eyes using the allocated width for the eyes


lukefromdc:
Fix
https://github.com/mate-desktop/mate-applets/issues/674

PR #678, opened 2025-06-03
 Allow to theme character picker buttons

luigifab:
Like https://github.com/mate-desktop/mate-panel/pull/1474

image

image

<del>But there is something wrong, for now I don't understand why the following CSS works only from GtkInspector!</del> -> `#PanelApplet`

```css
.mate-panel-menu-bar #PanelApplet .mate-charpick-applet .menu-button:checked  {
	color: green;
	border-color: blue;
	background:yellow;
}
.mate-panel-menu-bar #PanelApplet .mate-charpick-applet .text-button:active {
	/* border disabled by application */
	box-shadow:
		inset 1px 1px white,   /* top left */
		inset -1px -1px yellow; /* right bottom */
}
```

Maybe it's better to add the button menu in `curr_data`.

### DIFF
--- a/charpick/charpick.c
+++ b/charpick/charpick.c
@@ -270,10 +270,27 @@ key_press_event (GtkWidget   *widget,
 }
 
 static void
+menu_deactivated (GtkMenu   *menu,
+                  charpick_data *curr_data)
+{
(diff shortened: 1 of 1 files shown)


luigifab:
Ok working! CSS nodes are:

image

PR #679, opened 2025-06-30
invest: Bring back the invest applet

vkareh:
The original invest applet was dropped in
https://github.com/mate-desktop/mate-applets/commit/41f7b7a2a38ca9f3929fcbd88d228b3815034876 due to a broken API. This brings it back using the modern JSON version of the Yahoo Finance API, a rewrite from Python to C, and a simplified interface (i.e. instead of the full portfolio management app, I simplified it into a stock ticker price tracker).

A few files (like icons, help text, manpage, etc.) were pulled directly from git history, but I updated part of the help file to at least match what the new version of the applet does.

Screenshot at 2025-06-30 15-09-57
Screenshot at 2025-06-30 15-07-07
Screenshot at 2025-06-30 15-06-51


lukefromdc (commented):
The applet appears in the "add to panel" dialog but errors out of being added silently in an in-process build.The icon shown in the dialog is the "icon-missing" icon

Note that with an in-process build, the applet location in 
/usr/share/mate-panel/applets/org.mate.applets.InvestApplet.mate-panel-applet
was listed as 
`Location=/usr/libexec/invest-applet`
```
When it shoud have been 
Location=/usr/lib/x86_64-linux-gnu/mate-applets/libinvest-applet.so
```
for an install to
/usr/lib/x86_64-linux-gnu

but fixing just this wasn't enough to make the applet work. Same results both in x11 and wayland. we DID get the applet built as libinvest-applet.so but something failed badly enough that with the panel being run from terminal (with `mate-panel --replace`) I didn't even get any warning messages on trying to add it to the panel, just no response at all.

lukefromdc (approved):
Just rebuilt and retested. The applet now loads fine and shows its correct icon in the panel though I still got icon-missing in the panel add to dialog. I got "No valid stock data" thought maybe the endpoint was something I had .0.0.1'ed out in /etc/hosts

Turns out Yahoo's finance page is down at the moment, I fished the URL out of invest-applet.c and opened it in browser, getting the "Will be right back..." message.

With mate-panel running from terminal, this yielded 
`** (mate-panel:42020): WARNING **: 02:19:14.613: Failed to fetch stock data for symbol 0: Could not connect: Connection refused
`
when Yahoo could not take the more detailed URL due presumably to the outage. Presumably the applet is working as intended, though with this outage a more detailed check is not now possible.

Note that this is an in-process build-and it worked in wayland without any other issues

lukefromdc:
Let's see if we can get a second review for this

PR #680, opened 2025-07-22
invest: Add chart with stock price history

vkareh:
Adds a chart window that shows stock price history over a configurable time scale.

Screenshot at 2025-07-22 06-18-37

Screenshot at 2025-07-22 06-17-48

### DIFF
--- a/invest-applet/invest/Makefile.am
+++ b/invest-applet/invest/Makefile.am
@@ -11,6 +11,7 @@ AM_CPPFLAGS =			\
 
 APPLET_SOURCES =	\
 	invest-applet.c	\
+	invest-applet-chart.c	\
 	$(NULL)
 
 APPLET_LIBS =			\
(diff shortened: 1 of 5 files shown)


lukefromdc:
This seems to be the important part of the backtrace:
```

#0  0x00007ffff793e289 in gtk_widget_get_visible () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#1  0x00007fffef46ba62 in invest_chart_is_visible (chart=<optimized out>) at invest-applet-chart.c:151
#2  0x00007fffef469095 in invest_applet_destroy (widget=<optimized out>, applet=0x555555ba0ed0) at invest-applet.c:484
```

vkareh:
@lukefromdc thanks for catching that - I always forget to test things in-process :stuck_out_tongue: 
I'm having a hard time reproducing this consistently, but I've added some defensive NULL pointer checks and connected the destroy signal and it seems to fix it.

PR #689, opened 2026-03-11
netspeed: avoid glibtop warnings for non-existent network devices

vkareh:
When a network interface is disabled at the kernel level, the applet tries to read from a device that no longer exist, generating continuous warning messages in syslog.

Check if the device exists before calling glibtop_get_netload in both is_dummy_device and get_device_info. When the device doesn't exist, get_device_info returns a non-running device info, allowing the existing auto-change logic to find an active device.

Fixes #688

### DIFF
--- a/netspeed/src/backend.c
+++ b/netspeed/src/backend.c
@@ -64,10 +64,20 @@ struct nl80211_state {
 
 #endif /* HAVE_NL */
 
+static gboolean
+device_exists (const char *device)
+{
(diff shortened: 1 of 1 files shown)


mbkma (approved):
Good defensive fix. Checking that the interface still exists before asking glibtop for netload data should avoid the warning spam when a device disappears, and falling back to the non-running path keeps the auto-change logic working the way it already does.

PR #697, opened 2026-06-15
drivemount: fix use-after-free crash on icon theme change

### LINKED ISSUE #696: drivemount applet crashes (SIGSEGV) — use-after-free after a drive is removed and the icon theme changes
**Component:** `mate-drivemount-applet` (`drivemount/src/drive-button.c`)

### Summary

The Drive Mount applet crashes with SIGSEGV in its own code, dispatched from the
GTK main loop. It is a **use-after-free**: a `DriveButton` connects a `"changed"`
handler to the process-global default `GtkIconTheme` but never disconnects it on
destruction. After the button is freed (a volume/mount is removed), a later
icon-theme `"changed"` emission calls the handler on freed memory and schedules
`drive_button_update()` via `g_idle_add()`, which then crashes.

### Affected versions

Reproduced and observed on **1.26.1**. The same defect is present on current
**`master`** — `drive_button_dispose()` does not disconnect the handler.

### Frequency

On an affected machine (Linux Mint 22.3, mate-applets 1.26.1) `coredumpctl`
recorded repeated crashes over several months (2026-03-11, 03-16, 03-21,
04-11, 04-15, 05-05, 05-24, 06-15), typically around USB device plug/unplug
(which adds/removes drive buttons).

### Root cause

`drive_button_new()` / `drive_button_new_from_mount()` do:

```c
g_signal_connect (gtk_icon_theme_get_default (), "changed",
                  G_CALLBACK (drive_button_theme_change), self);
```

The default icon theme is a long-lived, process-global singleton.
`drive_button_dispose()` frees the button but never disconnects this handler, so:

1. A volume/mount is removed → the button is destroyed (`gtk_container_remove` → dispose).
2. A later `"changed"` emission invokes `drive_button_theme_change(theme, freed_self)`
   → `drive_button_queue_update(freed_self)` → `g_idle_add(drive_button_update, freed_self)`.
3. The idle fires `drive_button_update(freed_self)` →
   `g_return_val_if_fail (DRIVE_IS_BUTTON (freed_self), …)` dereferences the freed
   object's class pointer → **SIGSEGV**.

### Backtrace (stripped binary, from coredump)

```
#0  mate-drivemount-applet + 0x7093   (drive_button_update → DRIVE_IS_BUTTON check)
#3  g_main_loop_run
#4  gtk_main
```

The faulting instruction is `cmp %rsi,(%rax)` reading `obj->g_class->g_type` of a
freed `DriveButton` (its `g_class` word held garbage).

### AddressSanitizer confirmation

A minimal harness — create a `DriveButton` from a real mount, destroy it, emit
`"changed"` on the default icon theme, spin the main loop — reports:

```
ERROR: AddressSanitizer: heap-use-after-free   READ of size 4
  #0 drive_button_queue_update   drivemount/src/drive-button.c   (if (!self->update_tag))
  #1 drive_button_theme_change   drivemount/src/drive-button.c
  #5 g_signal_emit_by_name       ("changed")
  freed by:      g_object_unref     (button destroyed)
  allocated by:  drive_button_new_from_mount
```

With the handler disconnected in `dispose`, the same harness runs clean.

### Fix

Disconnect the handler in `drive_button_dispose()` — see linked PR.

Cigydd:
`DriveButton` connects `drive_button_theme_change` to the process-global default
`GtkIconTheme` `"changed"` signal in `drive_button_new()` and
`drive_button_new_from_mount()`, but `drive_button_dispose()` never disconnected
it. Because the default icon theme outlives the button, after a button is
destroyed (e.g. a volume or mount is removed) a later `"changed"` emission
invokes the callback on freed memory and schedules `drive_button_update()` via
`g_idle_add()`, crashing in the GTK main loop (SIGSEGV inside the
`DRIVE_IS_BUTTON()` type check).

On affected systems this applet has been crashing repeatedly for months, around
USB device plug/unplug.

Disconnect the handler in `drive_button_dispose()`:

```c
g_signal_handlers_disconnect_by_func (gtk_icon_theme_get_default (),
                                      G_CALLBACK (drive_button_theme_change),
                                      self);
```

Verified with AddressSanitizer: a harness that destroys a `DriveButton` and then
emits `"changed"` on the default icon theme reports `heap-use-after-free` in
`drive_button_queue_update` (via `drive_button_theme_change`) without this
change, and runs cleanly with it.

Fixes #696.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/drivemount/src/drive-button.c
+++ b/drivemount/src/drive-button.c
@@ -144,6 +144,15 @@ drive_button_dispose (GObject *object)
 {
     DriveButton *self = DRIVE_BUTTON (object);
 
+    /* The "changed" handler was connected to the process-global default icon
+     * theme (see drive_button_new / drive_button_new_from_mount), which
+     * outlives this button.  If we don't disconnect it, the theme keeps a
+     * dangling pointer to the freed button and a later "changed" emission
+     * schedules drive_button_update() on freed memory -> use-after-free crash. */
+    g_signal_handlers_disconnect_by_func (gtk_icon_theme_get_default (),
+                                          G_CALLBACK (drive_button_theme_change),
+                                          self);
+
     drive_button_set_volume (self, NULL);
 
     if (self->update_tag)


vkareh:
@Cigydd this change looks fine to me, but I would like your commit message attribution to follow MATE's contribution guidelines: https://github.com/mate-desktop/mate-applets?tab=contributing-ov-file#attribution

tl;dr: Clean up the commit message to be more readable, and change the co-authorship like this (or something reasonably similar to it):
```diff
- Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
+ Assisted-By: Claude:claude-opus-4.8
```

lukefromdc:
I looked over this thankfully very simple code change, basically the same as what a human author would have 
written. I guess the real question is why wasn't this included when the applet was first written? Or maybe it got
lost somewhere?

lukefromdc (commented):
This builds fine and stops the crash, which I was able to verify by first ejecting a drive with this applet in the panel, then changing icon themes. As expected, that took down the whole panel with all applets in-process.

Applying this change to disconnect the signal handler stops the crash.

Holding off on approval until the attribution is fixed, but if I wrote this code myself the only likely change would be the wording of the comment. I checked the docs for this over at
https://docs.gtk.org/gobject/func.signal_handlers_disconnect_by_func.html

Cigydd:
Thanks for the quick review and for verifying the crash, @lukefromdc!

I've cleaned up the commit message for readability and switched the attribution to `Assisted-by: Claude:claude-opus-4.8` per the contribution guidelines.

Re: why it wasn't there originally — the `g_signal_connect()` to the global default icon theme has been present since the applet's initial commit (312ba610), but a matching disconnect in `drive_button_dispose()` was never added. So it's a latent bug from day one; it only triggers on the specific "destroy a button, then change the icon theme" sequence, which is uncommon enough that it went unnoticed for years.

Cigydd:
Hi @lukefromdc,

Tried to fix the attribution. Is it OK now?

Dne 16. 06. 26 v 5:53 Luke from DC napsal(a):

lukefromdc:
Approving as the actual code while generated on an AI is the same thing I would have written by hand.

